PR #7, opened 2026-05-28
Implement setup and sync commands

nhorton:
## Summary
- implement Phase 4 setup, sync, and create_profile/create-profile command objects
- add URI profile source cache encoding with credential redaction for displayed/cacheable values
- update Phase 4 plan status and docs/README command examples

## Validation
- npm run check-ci
- npm run build
- git diff --check
- DeepWork reviews passed and marked

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ bridl run --profile engineering-default
 bridl run -p support -- --cwd ~/work/customer-issue
 bridl sync
 bridl setup
-bridl create_profile --scope user --name regulated
+bridl create_profile regulated --scope user
 ```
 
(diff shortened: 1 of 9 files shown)


nhorton:
Addressed Copilot review comments in 9d54871:

- setup now validates the full discovered settings stack before creating initial user files
- setup now creates the user default profile from the user settings/template instead of project-local/project defaults
- setup now validates the default profile id with the same filesystem-safe profile id rules before using it as a path segment
- added regression coverage for invalid project settings, project default isolation, and unsafe default profile ids

Validation:
- npm run check-ci
- npm run build
- git diff --check
- focused DeepWork reviews passed and were marked

PR #48, opened 2026-06-05
Add profile subcommands

nhorton:
## Summary
- replace top-level create_profile/create-profile with profile create
- add profile list for configured local and cached remote profile sources
- update tests, requirements, and docs for the profile command namespace

## Validation
- npm run check-ci


requirements/APPLEPI-REQ-004-sync-and-setup.md
```diff
@@ -35,16 +35,22 @@ ApplePi provides setup and maintenance commands that create initial configuratio
...
+1. ApplePi MUST provide a `profile create` command.
+2. The `profile create` command MUST require a destination scope or destination path.
+3. The `profile create` command MUST require a profile name.
+4. The `profile create` command MUST create a placeholder profile folder with a valid `profile.yml` file.
+5. The `profile create` command SHOULD create conventional subfolders for common profile resources.
 
 ### APPLEPI-REQ-004.4: Command Object Implementation
 
-1. The `setup`, `sync`, and profile-creation command entry points MUST execute command objects rather than duplicate implementation logic in parser callbacks.
+1. The `setup`, `sync`, and profile command entry points MUST execute command objects rather than duplicate implementation logic in parser callbacks.
```
nhorton:
```suggestion
1. All CLI command entry points MUST execute command objects rather than duplicate implementation logic in parser callbacks.
```
No reason to not generalize this.

nhorton: Done in 9029443. I generalized the requirement, and also split the profile namespace into separate Commander subcommand factories with addCommand().

src/cli/ApplePiCli.ts
```diff
@@ -11,7 +11,7 @@ export const createDefaultCommands = (): CommandObject[] => [
   createRunCommand(),
   createSetupCommand(),
   createSyncCommand(),
-  createCreateProfileCommand(),
+  createProfileCommand(),
 ];
```
nhorton, line 15: 1. It seems like this is missing the "profile list" command.
2. Maybe we should make the `createXCommand()` methods return arrays and then concat them here. You could then just import `createProfileCommand` here and it could return itself and all its children.

nhorton: Done in 1115635. `createDefaultCommands()` now spreads `createProfileCommands()`, which returns the profile parent plus `profile list` and `profile create` command objects. The child command objects register themselves under the shared `profile` Commander namespace.

PR #92, opened 2026-06-29
feat(profiles): generate pi subagents from profiles

ncrmro:
## Summary
- add top-level `agent_generation` profile setting
- generate runtime Pi `agents/<profile-id>.md` files for runnable profiles that opt in
- validate generated agents through profile loader, resolver, run command, and Pi adapter tests

## Validation
- `npm run lint --prefix outfitter`
- `npm run typecheck --prefix outfitter`
- `npm test --prefix outfitter`
- `npm run build --prefix outfitter`
- installed local Outfitter globally and validated `./link` leader run generates `agents/engineer.md` only`

PR #129, opened 2026-07-02
feat(profiles): generate Pi skills from profiles

ncrmro:
## Summary
- add `skill_generation: true` profile support that resolves marked profiles and writes generated Pi `SKILL.md` files into the composite profile
- enable generated profile skills in Pi launch args alongside the built-in Outfitter skill and profile-bundled skills
- document generated profile skills and cover schema parsing, generated profile resolution, and Pi adapter launch behavior in tests

## Validation
- `npm run typecheck`
- `npm run lint`
- `npm run coverage -- --coverage.reporter=text-summary`
- `npm run build`
- `/review` requested by user; DeepWork reported no matching review rules for the changed files

## Notes
- Checked subagents PR #92 first. It is still open with passing CI, but GitHub reports it is not currently mergeable because conflicts must be resolved locally before merge.

### DIFF
--- a/code/cli/src/agents/AgentAdapter.ts
+++ b/code/cli/src/agents/AgentAdapter.ts
@@ -15,6 +15,7 @@ export interface AgentLaunchProfileLayer {
   readonly profilePath: string;
   readonly sourceRootPath?: string;
   readonly resourceRootPath?: string;
+  readonly sourceInputs?: readonly string[];
   readonly layout?: 'directory' | 'flat-file';
 }
 
@@ -25,6 +26,7 @@ export interface AgentLaunchContext {
   readonly cacheDirectory?: string;
   /** Directory of bundled user-facing Outfitter documentation the launched agent may read. */
   readonly outfitterDocsDirectory?: string;
+  readonly generatedSkillProfiles?: readonly AgentLaunchProfileLayer[];
 }
 
 export interface AgentCompositeProfilePlan {
@@ -43,6 +45,7 @@ export interface AgentAdapter {
       readonly profilePaths: readonly string[];
       readonly profileFolders?: readonly string[];
       readonly profileLayers?: readonly AgentLaunchProfileLayer[];
+      readonly generatedSkillProfiles?: readonly AgentLaunchProfileLayer[];
       readonly homeDirectory?: string;
       readonly cacheDirectory?: string;
       readonly settings?: Settings;
--- a/code/cli/src/agents/pi/PiAdapter.ts
+++ b/code/cli/src/agents/pi/PiAdapter.ts
@@ -23,6 +23,8 @@ import type { StatePathDeclaration, CompositeProfileStatePath } from '../../comp
 import { createPiArgs } from './PiArgs.js';
 import { createPiMcpConfigFile } from './PiMcpConfig.js';
 import { materializePiExtensionSources } from './PiExtensionCache.js';
+import { createPiGeneratedSkillFiles, createPiGeneratedSkillSources } from './PiGeneratedSkills.js';
+import { createPiSkillSources } from './PiSkillSources.js';
 
 const piControlNames = new Set([
   ...[...genericControlNames].filter((controlName) => controlName !== 'pi' && controlName !== 'claude'),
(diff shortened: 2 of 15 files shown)


ncrmro:
Thanks for this — the plumbing is solid (resolution of inherited controls in `GeneratedSkillProfiles.ts`, the cycle/template guards and warnings, and extracting `PiSkillSources.ts`/`ProfileContributors.ts` are all improvements I'd want to keep regardless of direction). But before we ship `skill_generation`, I want to push back on the core design: converting a *profile* into a *skill* conflates two artifacts that I think should stay distinct.

### Progressive disclosure

Skills work because of progressive disclosure: only `name` + `description` sit in context at rest, and the body loads on invocation. The generated body in `PiGeneratedSkills.ts` mostly respects this today (it references prompt files rather than embedding them), but the shape pushes the wrong way:

- The default description (`Adopt the ${label} Outfitter profile.`) is a weak disclosure hook — the agent can't tell *when* to reach for it.
- The body tells the agent to read prompt files "when they are present in the current project or profile source", but `appendSystemPrompt` `file:` entries resolve against the declaring profile's source root, which isn't materialized into the launched session — so `prompts/founder.md` is often a dangling reference. The obvious fix is to inline the prompt text into the skill body, and that's exactly the front-loading failure `docs/philosophy.md` ("Composition over accumulation") warns about.
- Runtime hints like `- Provider:` / `- Model:` / `- Thinking:` are launch-time controls a running agent cannot honor; in a skill body they're noise.

### Skill-merge semantics

Two composition questions the PR leaves undefined:

1. Profiles bundle their own skills (`profiles/<id>/skills/`, `cli_specific/pi/skills/`). When a profile becomes a generated skill, those bundled skills are dropped — `createPiGeneratedSkillSources` only emits `skills/<profile-id>`, and `profileFolders` only covers the *selected* stack. Are a generated-skill profile's skills transitively merged, namespaced, or intentionally dropped? Whatever the answer, it needs documented semantics.
2. `findGeneratedSkillProfiles(loadedProfiles.profiles)` in `RunProfileResolution.ts` scans **all loaded profiles from all sources**, not the resolved stack. Any catalog author who sets `skill_generation: true` injects a skill into every session that loads that source, regardless of which profile the user selected. That inverts our precedence model: the consumer composes, the author doesn't push.

### Alternative: skills as first-class catalog artifacts

Instead of profiles *becoming* skills, make named skills peers of profiles in the catalog — id slug, versioned and shared the same way — and have profiles *reference* them:

```yaml
# catalog: skills/conventional-commits/skill.yml
id: conventional-commits
description: Format commits as type(scope) subject; one logical change per commit.
invocation: auto        # agent may invoke when relevant
# body in skills/conventional-commits/SKILL.md

# profile
id: reviewer
inherits: [base]
controls:
  skills:
    - conventional-commits
    - id: release-checklist
      invocation: user-only   # only the user can trigger it
```

What this buys us:

- **Reuse without duplication** — a founder-review skill is written once and referenced from any profile, instead of every profile that wants it being marked `skill_generation: true`.
- **Progressive disclosure by construction** — slug + description is the at-rest surface; the authored body loads on invocation. No temptation to inline system prompts.
- **Clean merge semantics** — skill lists compose by the same precedence rules as other controls: layers append, dedup by slug, higher-precedence layer wins per-slug overrides. No transitive-skill question, because a skill doesn't carry skills.
- **Adapter mapping is direct** — the invocation flag translates straight to Claude Code's existing frontmatter (`disable-model-invocation: true` for user-only, `user-invocable: false` for agent-only), and for Pi we materialize `skills/<slug>/SKILL.md` exactly like this PR already does. Today `controls.skills` takes paths and doesn't translate for Claude at all (support-matrix calls it out); slugs fix both at once.

If someone genuinely wants "adopt this profile's perspective mid-session", that's then just a skill in the catalog whose body says so — authored deliberately, not generated from launch config.

Happy to pair on reshaping this — most of the resolution and Pi materialization code here carries over almost unchanged.


---

**Addendum — the settings layer.** For completeness, here is how named skills land in `~/.outfitter/settings.yml`. No new source syntax is needed: a catalog source already contributes whatever it publishes, so `skills/` rides alongside `profiles/` in the same `profile_sources` entry. The only new surface (*proposed*, does not exist today) is a top-level `skills:` block for setting default availability and invocation policy, composing by the same layer-precedence rules as `controls.skills` above:

```yaml
# ~/.outfitter/settings.yml
default_profile: reviewer
profile_sources:
  - github: my-org/outfitter-catalog # existing syntax; source now contributes skills/ alongside profiles/
    ref: v1.2.0
  - path: ./profiles

# proposed: settings-layer skill defaults, merged below profile controls.skills
skills:
  - conventional-commits # available in every session, invocation: auto (from skill.yml)
  - id: release-checklist
    invocation: user-only # per-slug override at the settings layer
```

Profiles that list a slug in `controls.skills` win per-slug over these defaults, same append/dedup/override rules as everything else in the stack.

PR #152, opened 2026-07-10
docs: add external assets to the skill contract

ncrmro:
## Summary

Follow-up to #150: adds an `assets` frontmatter key to the skills contract for files the agent adapts to produce output (workflow templates, report skeletons) rather than reads for knowledge. It mirrors `references` exactly — same `file` / `repo_file` sources, validation, and trust rules, materializing under `assets/` instead of `references/`. No new per-entry fields.

First consumer: ai-outfitter/actions#13 ships its `examples/` workflows as skill assets.

## Validation

- `npx prettier --check docs/documentation/skills.md`
- `git diff --check`

### DIFF
--- a/docs/documentation/skills.md
+++ b/docs/documentation/skills.md
@@ -325,6 +325,19 @@ Each reference materializes as `references/<source basename>`. Two references
 whose sources share a basename fail validation; rename one of the source
 documents to resolve the collision.
 
+### External assets
+
+Files the agent adapts to produce output — workflow templates, report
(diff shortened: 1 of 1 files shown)


ncrmro:
Closing — the assets contract is being handled by another workstream. The proposed shape here (an `assets:` key mirroring `references:` — same file/repo_file sources, validation, and trust rules, materializing under assets/) is available in this PR's single commit if useful. ai-outfitter/actions#13 consumes the assets contract, whichever PR lands it.

Issue #149, opened 2026-07-10
Support catalog skills and external SKILL.md references


ncrmro:
## Problem

Outfitter profiles can select skill paths, but catalogs cannot expose skills as independently discoverable capabilities with stable IDs and external human-maintained references. Consumers often inherit a profile merely to receive one skill, duplicate documentation into a skill folder, or rely on fragile symlinks.

This pushes catalogs toward too many profiles and works against progressive disclosure.

## Proposed contract

- Derive a skill ID from its containing folder and require `SKILL.md` `name` to match.
- Resolve bare IDs from `controls.skills`, while retaining explicit paths for compatibility.
- Discover catalog-level `skills/` directories from configured profile repositories, including repositories that publish skills without placeholder profiles.
- Allow `SKILL.md` frontmatter to declare `references`.
- Resolve `references[].file` from the supplying catalog root.
- Resolve `references[].repo_path` from the active project root and treat it as untrusted.
- Materialize every selected reference beneath the generated skill's `references/` directory.
- Use the source basename by default and allow a basename-only `name` override.
- Default `required` to `true`; omit missing optional references.
- Do not interpolate reference contents or append them to the system prompt.
- Keep the feature adapter-neutral through top-level `controls.skills`.

## Why

Profiles should remain few and stable: identity, policy, tools, model controls, and write boundaries. Capabilities should grow through many focused skills that disclose instructions and references only when activated. Catalog users should not inherit an unrelated profile simply to consume a skill.

External references also let people and agents share canonical architecture and operating documents under ordinary `docs/` paths without copying them into skill packages.

## Trust and validation

- `file` references inherit the trust of their catalog skill.
- `repo_path` references are untrusted consuming-repository content.
- Resolved targets must remain within the catalog or project root after following symlinks.
- Broken, escaping, non-file, duplicate, and ambiguous required references should fail lint/launch validation.
- Shadowed skill IDs should produce a visible diagnostic while following deterministic source precedence.

## Acceptance criteria

- [ ] Profile and skill schemas support the documented contract.
- [ ] Bare skill IDs resolve from local and cached remote catalog roots.
- [ ] Catalog skills can be consumed without profile inheritance.
- [ ] References materialize under generated skill `references/` directories.
- [ ] Pi and Claude adapters receive the same resolved top-level skills where supported.
- [ ] `outfitter profile lint` diagnoses unresolved IDs and invalid references.
[...]

PR #157, opened 2026-07-10
docs: allow glob targets in skill references

### LINKED ISSUE #156: Support glob targets in skill reference and asset entries
## Problem

The skills contract (docs/documentation/skills.md, from #150) defines `file` / `repo_file` reference entries as single-file targets. Catalog skills that want to expose an entire docs directory must re-list every file and edit frontmatter each time a doc is added.

ai-outfitter/actions#13 is a waiting consumer: its `outfitter-actions` skill declares `file: docs/*.md` (references) and `file: examples/*.yml` (assets) so new human-maintained docs and workflow templates flow into the skill without frontmatter edits.

## Proposed contract

- A `file` or `repo_file` target may be a glob, expanded at resolution time against its root (catalog root for `file`, active project root for `repo_file`).
- Each match materializes under the entry's destination directory by basename, exactly as if listed individually; basename collisions fail validation as today.
- A `file` glob matching zero files fails validation (consistent with a broken `file` target). A `repo_file` glob matching zero files is omitted (consistent with a missing `repo_file` target).
- Matches must stay within their root after symlink normalization, per the existing trust rules; recursive `**` follows the same constraint.
- Applies to `assets` entries identically if/when that key lands (see the closed #152 proposal).

## Scope

Docs-first, same as #150: define the semantics in skills.md; runtime lands with #149.

ncrmro:
Extends the skills reference contract so `file` and `repo_file` targets may be globs, expanded at resolution time against their respective roots. Matches materialize under `references/` by basename with the existing collision rule; a zero-match `file` glob fails validation while a zero-match `repo_file` glob is omitted, and matches must stay within their root after symlink normalization (`**` included). The same semantics apply to `scripts` and `assets` entries.

Docs-first: this documents the contract only; runtime implementation is tracked in #149. ai-outfitter/actions#13 is waiting on this behavior.

Closes #156

Validation:
- `npx prettier --check docs/documentation/skills.md`
- `git diff --check`

### DIFF
--- a/docs/documentation/skills.md
+++ b/docs/documentation/skills.md
@@ -325,6 +325,25 @@ Each reference materializes as `references/<source basename>`. Two references
 whose sources share a basename fail validation; rename one of the source
 documents to resolve the collision.
 
+### Glob targets
+
+A `file` or `repo_file` target may be a glob, expanded at resolution time
+against its root — the profile-repository checkout for `file`, the active
+project root for `repo_file`:
+
+```yaml
+references:
+  - file: docs/*.md
+```
+
+Each match materializes under `references/` by basename, exactly as if listed
+individually, and joins the same collision check. A `file` glob matching zero
+files fails validation, like a broken `file` target; a `repo_file` glob
+matching zero files is omitted, like a missing `repo_file` target. Matches
+MUST remain within their root after symlink normalization, per
+[Trust boundary](#trust-boundary); `**` is allowed under the same constraint.
+The same semantics apply to `scripts` and `assets` entries.
+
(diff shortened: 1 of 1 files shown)


ncrmro:
Superseded by #158, which unifies this glob-target contract with directory-target support into one spec (file | directory | glob for references, scripts, and assets) and implements it in SkillResolution.ts with tests — including the zero-match semantics (file fails, repo_file omitted), per-match symlink-normalization escape checks with `**`, and the basename collision rule specified here. The docs wording from this PR was incorporated into docs/documentation/skills.md there.

PR #159, opened 2026-07-11
feat(run): export composed Claude system prompt on profile_export

ncrmro:
## Problem

`profile_export: true` claims to write the generated system prompt beside the selected profile, but it only ever worked for **Pi**:

- `renderGeneratedSystemPrompt` hard-coded the Pi rendering and merged only `pi` controls.
- It filtered append entries down to plain strings, so typed `{ file: … }` / `{ repo_file: … }` includes were **silently dropped** and rendered as `(none)`.
- The seeded artifact was only made real by the Pi launch extension's `ctx.getSystemPrompt()` overwrite.

Claude Code has no runtime prompt read-back, so a Claude launch left the export an empty stub (`system_prompt (none)` / `append_system_prompt (none)`) even when the profile defined prompt includes.

## Change

Make the export adapter-aware and reuse the launch-time `resolveAppendSystemPromptControl`, so the artifact reflects the exact composed prompt each adapter actually receives — with no launch flag beyond selecting the adapter:

- **Pi**: output is byte-identical to before (the locked OFTR-005.7 tests are untouched and still pass) and is still overwritten at session start by the runtime capture.
- **Claude**: the export records the effective `system_prompt` and the **resolved** `append_system_prompt` includes Outfitter hands Claude via `--system-prompt` / `--append-system-prompt`, with a header that's honest about it reflecting Outfitter's supplied inputs rather than Claude's full base prompt.

`agentId` + `projectDirectory` are threaded from `RunCommand` into the exporter (both the initial export and the watch-refresh). Requirement OFTR-005.7 and the profile docs are generalized from "Pi prompt export" to "system prompt export".

## Tests

- All existing OFTR-005.7 export tests pass unchanged (Pi output byte-identical).
- Added two Claude tests: one asserting typed file includes resolve to their contents (not `(none)`) with the Claude header and no extra flags, and one asserting `claude:`-specific prompt overrides win.
- Full suite green: **360 passed**; typecheck + eslint + prettier clean.

## Manual verification

Built via `bin/dev-install-outfitter-dev` and ran `outfitter run -p researcher --agent claude`. The generated `researcher.generated-system-prompt.md` now contains all three resolved context includes instead of `(none)`.


ncrmro:
Closing as stale. The RFC #165 dotagents refactor (merged in #172–#182) replaced the profile system this PR builds on with the `.agents` resolve→compose→project pipeline. Composed system-prompt export is now covered by `outfitter dump`. The branch is retained if we want to revisit.

Issue #220, opened 2026-07-26
Follow-ups from remote source sync review (#219)


ncrmro:
Non-blocking findings deferred from the review of #219. The blocking security and resolution findings were fixed in that PR (`e746388`); these remain.

## Temporary directory and XDG layout

`outfitter run` uses `mkdtemp(tmpdir(), 'outfitter-<agent>-<harness>-')` as `PI_CODING_AGENT_DIR` (`RunAgentCommand.ts:179`) and removes it in a `finally`. Measured on one developer machine over ~7 days: **32 stranded directories, 2.2MB**, roughly 4-5 leaked per day. The `finally` cannot run on Ctrl-C — SIGINT reaches the whole foreground process group and there are no signal handlers.

Two consequences worse than the disk usage:

1. **Stranded credentials.** 7 of those directories contain `auth.json`. On a typical Linux box `/tmp` is real disk with a 10-day `systemd-tmpfiles` sweep, so credentials land on the filesystem (and in any snapshot) rather than tmpfs.
2. **Session history is destroyed on clean exit.** pi writes `sessions/<project>/<ts>_<uuid>.jsonl` into the agent dir — observed up to 2.0MB. `rmSync` deletes the transcript on normal exit; it only survives when the CLI crashes, stranded in `/tmp`. That is backwards.

The agent directory currently mixes three lifetimes. Proposed split:

| Content | Now | Proposed |
|---|---|---|
| `agent.md`, `system-prompt.md`, `settings.json`, pi overlays | `/tmp` | `$XDG_RUNTIME_DIR/outfitter/run-<pid>-<nonce>` (tmpfs; falls back to `tmpdir()` where unset, e.g. macOS) |
| `sessions/` | `/tmp`, deleted on exit | symlink to `$XDG_STATE_HOME/outfitter/sessions/<agent>` so transcripts persist and are resumable |
| `auth.json`, `models-store.json` | copied to `/tmp` and back | symlink into the pi user agent dir |

Leak cleanup should be a **startup sweeper** (prune sibling `run-*` dirs whose pid is dead or that exceed an age bound), not signal handlers — handlers still miss SIGKILL, OOM, and terminal death, so the crash path can never be covered by `finally`.

Already correct and out of scope: pi extensions (`~/.cache/outfitter/pi-extensions` via `resolveOutfitterCacheDir`) and git fetch temporaries (siblings of the cache path under `~/.agents/cache/repos`).

## Smaller items

[...]

PR #222, opened 2026-07-26
docs(architecture): specify projection runtime layout and cleanup

ncrmro:
Companion to [State Writeback Strategy](../blob/main/docs/architecture/state_writeback_strategy.md), which specifies what happens to writes *inside* a projection but assumes two things it never states: that a projection root exists somewhere appropriate, and that its "clean up temporary state" step actually runs. Both assumptions fail observably today.

## What it specifies

- **Two cache roots.** `<cache_directory>` (the OFTR-004 `.agents` source cache, default `~/.agents/cache`) and the Outfitter XDG cache root (`~/.cache/outfitter`) are unrelated. Conflating them is easy and this document does it first so nothing else has to.
- **Lifetime classes** and the rule for classifying a new file — projection, native state, build artifact, source cache.
- **Runtime root** resolution (`$XDG_RUNTIME_DIR` first, so credentials reached through a projection never touch persistent storage), naming, mode, per-segment path length bounds, and how the root reaches the child via `PI_CODING_AGENT_DIR` / `CLAUDE_CONFIG_DIR`.
- **Cleanup** as a single pid-liveness precedence rule. Exit handlers cannot be made reliable: `SIGINT` reaches the whole foreground process group, so interrupting the agent interrupts Outfitter at the same instant.
- **Extension cache** ownership, why installs stay out of the native agent directory, and the missing reclamation path.
- **Security invariants**, concurrency, and the platform matrix.

## Divergences recorded

Five, the largest being that `state_persistence` is parsed and merged but never consumed and no adapter implements the declared `state_paths` — so the declared paths are unimplemented design, not broken behavior. The document says so explicitly to avoid reading like a bug list.

One divergence is new information: **`models-store.json`** is written by pi into the agent directory, appears in no adapter `state_paths` declaration, and is not in the copied set (`['auth.json', 'models.json']`), so it is discarded every run. It is either native state that should be declared or a cache file that should be named as such.

The `/tmp` leak measurement is dated (seven days ending 2026-07-25: 32 stranded roots, 7 containing `auth.json`) so the claim can be aged.

## Also in this PR

`docs/architecture/README.md` documented `$TMPDIR/outfitter-<agent>-<harness>-<random>/` as target state, which this document contradicts. Reconciled and linked, so the new document is not orphaned.

## Notes for review

- **No RFC 2119 keywords.** Per OFTR-008, normative statements belong in numbered `OFTR-NNN` files where a pinned test can reference them. The `## Security invariants` section is the strongest candidate for promotion and says so.
- **One open question is genuinely open**: whether a given harness tolerates concurrent writes to its session store and auth file. That needs an experiment per adapter, not a design decision, and is marked as such rather than assumed.

Follows up #220.

### DIFF
--- a/docs/architecture/README.md
+++ b/docs/architecture/README.md
@@ -158,11 +158,13 @@ Generated harness files may use LiquidJS templating with Outfitter-specific deli
 
 ## Runtime Model and State
 
-A launch materializes the composed agent under the system temp directory:
+A launch materializes the composed agent under a per-user runtime directory:
 
 ```text
-$TMPDIR/outfitter-<agent>-<harness>-<random>/
+$XDG_RUNTIME_DIR/outfitter/outfitter-<agent>-<harness>-<pid>-<nonce>/
 ```
+
+Projection location, cleanup, and the cache roots are specified in [`./projection_runtime_layout.md`](./projection_runtime_layout.md), which also records where current behavior still uses the system temp directory.
 
 Adapter-declared state paths symlink to durable native CLI locations (`~/.pi/agent/...`, `~/.claude/...`) per the functional state model — see [`./state_writeback_strategy.md`](./state_writeback_strategy.md). State handling strategies are `symlink`, `discard`, `warn`, `error`, and `prompt`; unknown writes are never silently persisted. `state_persistence` policy lives in settings.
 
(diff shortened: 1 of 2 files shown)


ncrmro:

Issue #226, opened 2026-07-27
Ship the persona review journey


ncrmro:
Shipped one fresh-user path:

```text
outfitter setup
→ outfitter run founder
→ persona-authoring writes one portable Markdown persona
→ persona-review launches one isolated outfitter run persona-reviewer process
→ a sourced report is saved under docs/persona-reviews/
```

The reviewer is a normal process launched by the shared skill. This journey
does not depend on Pi's native subagent projection.

## Review and release order

- [x] Review and merge ai-outfitter/community-profiles#21.
- [x] Merge ai-outfitter/community-profiles#20 and publish
      `community-profiles@v1.1.1`.
- [x] Review and merge ai-outfitter/default-profiles#17.
- [x] Merge ai-outfitter/default-profiles#14 and publish
      `default-profiles@v1.1.0`.
- [x] Review and merge #227.
- [x] Review and merge #228 and publish Outfitter v1.1.1.

## Acceptance

- A fresh `outfitter setup` resolves the released founder and persona
  resources through immutable catalog pins.
- `persona-reviewer` inherits the caller's configured model instead of
  requiring provider-specific credentials.
- Synced consumers can run the reviewer through the direct `outfitter run`
  command without locating a cache-relative launcher.
- The `persona-review` skill can dispatch the finite reviewer in the
  background when `interactive_shell` is available, fall back to a synchronous
  shell, and atomically save a non-empty report at the requested durable path.
- The founder invokes that skill instead of native subagent delegation.
- Platform engineers can distribute the shared capability to both engineer
  and marketing profiles, whose loadouts explicitly select `persona-review`.

## Separate follow-ups

- Native harness subagent projection remains #183.
- Source-add ergonomics, a recorded demo, and benchmark/eval coverage are
  useful follow-ups, not release blockers for this journey.

ncrmro:
Released end to end: community-profiles v1.1.1, default-profiles v1.1.0, and Outfitter v1.1.1. The npm package and versioned/latest GHCR images were published successfully.

PR #227, opened 2026-07-27
fix(setup): ship the persona review journey

ncrmro:
## Summary

- update the built-in immutable catalog pin to `default-profiles@v1.1.0`
- update exact pin coverage
- teach the portable direct `outfitter run persona-reviewer` command first
- document durable report capture and the optional skill-owned process wrapper
- state explicitly that native Pi subagent projection is not required

## Validation

- `npm run check-ci` passes
- 252 tests across 25 files pass
- documentation lint and production build pass
- the packaged CLI end-to-end smoke test passes

## Release chain

Depends on ai-outfitter/default-profiles#17 and publication of
`default-profiles@v1.1.0` through ai-outfitter/default-profiles#14. This is the
final implementation step for #226 and should produce the next Outfitter patch
release after it merges.

Source-add ergonomics, demo work, benchmarks, and native subagent projection
remain separate follow-ups.

### DIFF
--- a/code/cli/src/setup/DefaultCatalog.ts
+++ b/code/cli/src/setup/DefaultCatalog.ts
@@ -10,7 +10,7 @@ export const defaultCatalogSource = {
   github: 'ai-outfitter/default-profiles',
   // TODO(release-automation): when default-profiles publishes a new Release Please tag, open a
   // conventional dependency-bump commit in Outfitter so its next Release Please release ships it.
-  ref: 'v1.0.0',
+  ref: 'v1.1.0',
 } as const satisfies RemoteSourceReference;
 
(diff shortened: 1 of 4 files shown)


ncrmro (commented):
No actionable defects remain on the final head. Review corrected the organization diagram so its engineer and marketing profiles explicitly select the shared skill, matching Outfitter's loadout semantics. I reviewed the complete pin, test, persona-spec, and use-case diff. `npm run check-ci` passes with 252 tests across 25 files, lint, documentation lint/build, and packaged E2E coverage. Merge remains ordered behind publication of `default-profiles@v1.1.0`. This is a COMMENT rather than an approval because the authenticated account authored the PR.

Issue #234, opened 2026-07-28
feat: make the published container a reusable resident-agent base


ncrmro:
## Summary

Make the published Outfitter container directly usable as a persistent agent
runtime and straightforward to extend with Nix packages, without requiring
consumers such as Link Operator to publish a custom agent image.

The current image is Nix-built but intentionally minimal: it contains the
Outfitter closure and uses `outfitter` as its entrypoint. It does not expose
`/bin/sh` or the `nix` CLI, and invoking the image without an Outfitter
subcommand does not start a resident agent.

Related consumer work:

- ai-outfitter/link-operator#13
- ai-outfitter/link-operator#14

### Run the published image directly

As a platform engineer, I want to use the published Outfitter image directly as
an agent runtime so that an operator does not need to rebuild or republish
Outfitter.

- [ ] Define a supported persistent/resident container invocation.
- [ ] The resident process starts an Outfitter-selected agent, stays idle until
      work arrives, and handles termination signals.
- [ ] Profile identity and settings are supplied through mounted configuration
      and environment variables rather than baked into the image.
- [ ] The runtime does not bundle Slack, relay, MCP, or other channel-specific
      behavior.
- [ ] Document the Kubernetes container contract, including entrypoint,
      arguments, working directory, home directory, writable paths, and
      non-root execution.

### Extend the image with Nix packages

As a platform engineer, I want to derive a user-owned image with additional
tools so that each agent can carry its own runtime dependencies without changing
Outfitter or its operator.

- [ ] Document that "Nix-built" does not imply that the runtime image contains
      the `nix` CLI or a shell.
- [ ] Provide a supported Nix example that consumes
      `outfitter.packages.<system>.outfitter` and adds packages such as
      `pkgs.nix`, Bash, coreutils, certificates, Git, and user-selected tools
      through `dockerTools.buildLayeredImage`.
- [ ] Consider exposing a small reusable flake helper for assembling an
      Outfitter image from the Outfitter package plus caller-selected packages.
- [ ] The example image runs as a non-root UID and documents any writable
      `/nix` store/state requirements.
- [ ] Smoke test both `outfitter --version` and the added runtime tools inside
      the built image.

### Apply profile changes dynamically

As a platform engineer, I want changes to mounted profile/settings sources to
apply automatically so that updating an Agent custom resource does not require
rebuilding its image.

- [ ] `outfitter run` can detect and apply a changed profile/settings
      projection, or exposes a documented reload signal/command.
[...]

PR #235, opened 2026-07-28
feat: make the container a reusable agent runtime

ncrmro:
## Summary

- make the published image a channel-agnostic resident-agent base with Nix, Bash, core utilities, Git, SSH, and CA certificates
- export `lib.mkContainer` so users can reproducibly add their own Nix packages without Link Operator publishing a custom agent image
- run as a resolvable non-root UID/GID 1000 and document the Kubernetes, profile, credential, extension, and writable-Nix-state boundaries
- extend the release smoke test to cover Nix and conventional shell paths

This is the first usable vertical slice of #234. Dynamic profile hot reload remains follow-up work because the current runtime does not expose a clean reload hook.

## Validation

- `npm run check-ci` (Prettier, ESLint, 294 tests, and coverage passed; the docs build was rerun outside the filesystem sandbox because Turbopack binds an internal local port)
- `nix flake check --no-build`
- built `.#container` with the official Nix image
- loaded the result into Docker and verified the default non-root user, `nix --version`, Bash, `/usr/bin/env`, Git, SSH, and `outfitter --version`

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -91,6 +91,60 @@ jobs:
           docker run --rm --entrypoint /bin/sh "$IMAGE:$VERSION" -c \
             'nix --version && test -x /bin/bash && test -x /usr/bin/env'
 
+      # The two workflows this image exists to support. Neither was covered:
+      # the checks above only prove tools exist in the base, so a regression in
+      # extension or residency could still be published.
+      - name: Smoke test derivative image
+        run: |
+          set -euo pipefail
+          # A consumer extends the base with its own entrypoint. This failed
+          # before /etc/passwd carried root: the builder resolves USER against
+          # the base image's passwd file, so `USER root` aborted the build with
+          # "unknown user error looking up user root".
+          workdir="$(mktemp -d)"
+          printf '#!/bin/bash\nexec outfitter "$@"\n' > "$workdir/entrypoint"
+          cat > "$workdir/Dockerfile" <<DOCKERFILE
+          FROM $IMAGE:$VERSION
+          USER root
+          COPY entrypoint /opt/agent/entrypoint
+          RUN chmod 0755 /opt/agent/entrypoint
+          USER 1000:1000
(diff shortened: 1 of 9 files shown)


ncrmro:
Rebased onto current `main` (was based on 2026-07-28) and verified against a real consumer, plus one fix.

## Fixed: the base could not actually be used as a base

`/etc/passwd` carried only the `outfitter` user. A derived image doing `USER root` to copy in its own entrypoint failed before executing anything:

```
determining run uid: user: unknown user error looking up user "root"
```

The builder resolves `USER` against the base image's passwd file, so every downstream `RUN` failed. Added `root` and `nobody` entries. Reproduced first, then fixed.

## Verified

Built `.#container` from this branch and loaded it:

- `/bin/sh` works — this is the #234 defect; the published `1.3.1` image has no shell at all, so `FROM` + `RUN` is impossible against it today.
- `outfitter --version` → `1.3.1`, `git --version` → `2.54.0`.
- Built a **derived image** the way a consuming org would (`FROM` + `USER root` + `COPY entrypoint` + `RUN chmod`) — succeeds.
- That derived image resolves a real catalog profile and projects its MCP server into the composed `.agents/mcp.json`, so the extension path works for the case that motivated it: a resident agent whose profile declares `mcp:`.

## Why this matters beyond convenience

Consumers currently fork the runtime image because they cannot extend this one. ai-outfitter/agent-operator#13 wants the operator to stop publishing an agent image and default `Agent.spec.image` here instead; that is blocked until `FROM` works. One downstream has already published an agent image from an application repo purely to get a shell and a tool-restricting entrypoint.

## Not addressed here

The image is still large (~857 MB; a derived one ~891 MB) because the runtime set includes `nix`, `openssh`, and friends. A minimal `node:alpine` + `npm i -g @ai-outfitter/outfitter` equivalent is ~334 MB. Worth a follow-up on what belongs in a base versus what a consumer adds.

Issue #187, opened 2026-07-20
Track deferred artifact baking and managed harness projection


ncrmro:
Follow-up to #165.

## Summary

Defer persistent artifact baking and managed harness filesystem projection from the initial Dotagents cutover. The initial refactor keeps the resolver, composition, validation, deterministic `dump`, and temporary per-run projection required by `outfitter run`; it does not install persistent files or symlinks into harness-owned directories.

This issue owns the later design and implementation of persistent artifacts and opt-in, managed projection into native Pi, Claude Code, Codex, and future harness locations.

## Current boundary

- `outfitter run` materializes a temporary runtime directory and removes it after the child harness exits.
- `outfitter dump` writes a deterministic reviewable `.agents` tree, but is not a content-addressed bake and installs nothing.
- `outfitter setup` writes initial `.agents` settings and resources; it does not currently manage harness symlinks.
- Existing user-managed links and harness configuration must remain untouched during the initial refactor.
- Documentation that currently claims Claude configuration is ported and symlinked must be corrected to describe this as deferred work.

### Artifact baking

- Define the boundary between deterministic dumps, reusable content-addressed artifacts, and task-specific execution artifacts.
- Define artifact identity, provenance, storage, review/export behavior, garbage collection, and safe handling of secrets and mutable harness state.
- Keep task schemas and task execution in their own future RFC while allowing that work to reuse the artifact substrate designed here.

### Managed harness projection

- Treat `~/.agents/AGENTS.md` as canonical personal/global guidance and map it into harness-global instruction locations that support it.
- Treat `<project>/.agents/AGENTS.md` as an ordinary directory-scoped `AGENTS.md`; it is not composed or injected as runtime context.
- Do not give lowercase `.agents/agents.md` any special meaning.
- Define global and project mappings for Pi, Claude Code, Codex, and future harnesses, including instruction files, skills, agents/subagents, MCP, commands, models, and other supported resources.
- Prefer managed leaf symlinks, with an explicit copy fallback for platforms or consumers where symlinks are unsuitable.
- Require opt-in execution, dry-run output, an ownership manifest, conflict preservation, idempotent reconciliation, rollback/uninstall, and cross-platform behavior.
- Never overwrite, adopt, delete, or rewrite an unmanaged existing harness file or symlink silently.
- Keep persistent installation separate from the temporary projection lifecycle used by `outfitter run`, even if both share adapter mapping code.

## Related deferred work

[...]

PR #243, opened 2026-07-31
feat(pi): persist sessions across runs by default

### LINKED ISSUE #223: Persist Pi sessions by default for interactive outfitter run
## Problem

`outfitter run` currently launches Pi with `PI_CODING_AGENT_DIR` pointing at a per-run temporary projection directory. Pi writes session transcripts under that directory by default, and Outfitter deletes the projection after the process exits. As a result, `pi --continue` / resume cannot find the previous session on the next `outfitter run`.

In a live run, the active session was under a temp projection:

```text
PI_CODING_AGENT_DIR=/tmp/outfitter-founder-pi-EI5bXu
/tmp/outfitter-founder-pi-EI5bXu/sessions/--home-ncrmro-repos-ai-outfitter--/*.jsonl
```

The durable default Pi agent directory only contained an older unrelated session:

```text
~/.pi/agent/sessions/--tmp-tmp.kZZe8zSNvD--/*.jsonl
```

This makes session resume effectively broken for normal interactive Outfitter usage.

## Expected behavior

Interactive `outfitter run` SHOULD preserve Pi session history by default so users can resume work across launches.

Outfitter SHOULD keep the runtime projection disposable, but MUST NOT place the default interactive session store exclusively inside that deleted projection.

## Proposed behavior

- For interactive Pi launches, Outfitter SHOULD set a durable session directory, either via `PI_CODING_AGENT_SESSION_DIR` or `--session-dir`.
- The default target SHOULD be stable and project-aware, for example Pi's native durable session location or an Outfitter-owned XDG state path.
- Non-interactive/headless/CI launches MAY continue to discard sessions or use `--no-session`, but that SHOULD be explicit and documented.
- Users SHOULD be able to opt into ephemeral sessions with `--no-session`, a future CLI flag, or `state_persistence: { "sessions/": discard }` once state persistence is wired for this path.

## Related resident-agent footgun

`link-operator/code/agent/entrypoint.sh` on main currently launches Pi with `--no-session`, which explicitly disables persistence. The agent-session-gateway worktree changes this directionally correctly by using `--continue` plus a PVC-backed `PI_CODING_AGENT_SESSION_DIR=/workspace/.pi/agent/sessions`.

Resident agents MUST use a workspace/PVC-backed session directory when conversation continuity is required.

## Acceptance criteria

- [ ] Interactive `outfitter run` creates/resumes Pi sessions from a durable directory by default.
- [ ] A second `outfitter run -- --continue` in the same project can see the prior session after the first projection directory has been removed.
- [ ] Non-interactive modes retain an explicit, documented policy for whether sessions are persisted or discarded.
- [ ] Tests cover the launch plan/session-dir behavior for Pi.
- [ ] Documentation explains the default and how to opt out for privacy/reproducibility.

ncrmro:
Closes #223.

## Problem

`outfitter run` launches pi with `PI_CODING_AGENT_DIR` pointed at a per-run projection root under the system temp directory, and `RunAgentCommand` deletes that root once the child exits. Pi stores transcripts under `<agent dir>/sessions/<encoded cwd>` (`pi-coding-agent` `core/session-manager.js: getDefaultSessionDirPath()`), so every session was destroyed with the projection and `pi --continue` / `--resume` could never find the previous conversation.

## Change

Default `PI_CODING_AGENT_SESSION_DIR` to pi's own durable per-project session folder, `~/.pi/agent/sessions/--<encoded cwd>--`.

Why that path rather than the sessions root:

- When pi is given an explicit session directory it writes **flat** into it and switches `--continue`/`--resume` to a cwd filter that reads every transcript header in the directory. Pointing at `~/.pi/agent/sessions` would pile every project into one folder and put resume on the slow path.
- The per-project folder is the exact directory a standalone `pi` uses in that project, so history is shared both ways and pi keeps its fast path (`dir === getDefaultSessionDirPath(cwd)`).

Nothing is copied back after the run — pi writes straight to the durable directory, unlike the `auth.json`/`models.json` seed-and-write-back in `PiCredentialPersistence`.

Escape hatches, all honoured by pi's own precedence (`--session-dir` → `PI_CODING_AGENT_SESSION_DIR` → `sessionDir` in settings.json):

- `outfitter run -- --no-session` — record nothing.
- `outfitter run -- --session-dir <dir>` — choose the store per run.
- An inherited `PI_CODING_AGENT_SESSION_DIR` is left untouched, so a resident agent's workspace/PVC path keeps winning (the footgun called out in #223).

The default applies to interactive and non-interactive pi launches alike; a split policy would make print-mode sessions quietly unfindable. Claude Code is unchanged — `CLAUDE_CONFIG_DIR` session state is a separate gap (OFTR-006.5.6, #187).

## Verification

Unit tests cover the path encoding, the inherited-value opt-out, the projected launch env for pi vs claude, and a two-run `executeRunAgentCommand` case asserting both transcripts survive after the first projection is removed.

Verified end-to-end against the bundled pi 0.80.3 with an isolated `HOME`:

- `outfitter run engineer -- -p "reply with exactly: alpha"` → transcript written to `<home>/.pi/agent/sessions/--<project>--/…jsonl`, projection removed.
- `outfitter run engineer -- --continue -p "what single word did you reply with a moment ago?"` → answered `alpha`, appending to the same session file.
- With `PI_CODING_AGENT_SESSION_DIR` exported, the session landed in the override directory and the durable folder was untouched.

## Acceptance criteria (#223)

- [x] Interactive `outfitter run` creates/resumes pi sessions from a durable directory by default.
- [x] A second `outfitter run -- --continue` in the same project sees the prior session after the first projection is removed.
- [x] Non-interactive modes have an explicit, documented policy (same durable default).
- [x] Tests cover the launch plan / session-dir behavior for pi.
- [x] Documentation explains the default and how to opt out.

Docs: new "Pi sessions" section in `docs/documentation/state.md`, pi adapter note in `docs/architecture/README.md`, and a new requirement OFTR-006.3.19.

### DIFF
--- a/code/cli/src/agents/PiSessionDirectory.ts
+++ b/code/cli/src/agents/PiSessionDirectory.ts
@@ -36,7 +36,8 @@ export const resolvePiUserSessionDirectory = (homeDirectory: string, projectDire
  * Session directory for a pi launch, or `undefined` when the launch should keep whatever
  * `PI_CODING_AGENT_SESSION_DIR` it inherits — an explicitly configured store, such as a resident
  * agent's workspace/PVC path, stays in control. `env`, `homeDirectory`, and `projectDirectory` are
- * injected so the resolution stays pure and unit-testable.
+ * injected so the resolution stays unit-testable; `projectDirectory` is the directory pi runs in
+ * (`process.cwd()`), which is what makes the result pi's own folder for the project.
  */
 export const resolvePiSessionDirectory = (
   env: Readonly<Record<string, string | undefined>>,
--- a/code/cli/src/agents/PiSessionDirectory.ts
+++ b/code/cli/src/agents/PiSessionDirectory.ts
@@ -1,0 +1,50 @@
+// Keeps pi session transcripts durable across runs. Outfitter launches pi with PI_CODING_AGENT_DIR
+// pointed at a per-run projection root that is deleted once the child exits, and pi stores sessions
+// under `<agent dir>/sessions/<encoded cwd>` by default (pi-coding-agent core/session-manager.js:
+// getDefaultSessionDirPath()). Left alone every transcript dies with the projection, so
+// `pi --continue`/`--resume` can never find the previous conversation (#223).
+//
+// Outfitter therefore points pi at the same per-project folder under pi's durable agent directory
+// that a standalone pi already writes to, so history is shared between `outfitter run` and a bare
+// `pi` in the same project, and `--continue` keeps pi's fast path instead of scanning every
+// project's transcripts.
+import { join, resolve } from 'node:path';
+
+import { resolvePiUserAgentDirectory } from './PiCredentialPersistence.js';
+
+/**
+ * pi's session-storage environment override. pi resolves the session directory as
+ * `--session-dir` → `PI_CODING_AGENT_SESSION_DIR` → `sessionDir` in settings.json, so a
+ * pass-through `-- --session-dir <dir>` still wins over the default Outfitter sets here.
+ */
+export const PI_SESSION_DIRECTORY_ENV = 'PI_CODING_AGENT_SESSION_DIR';
+
+// pi's own encoding of a project directory into a single session folder name: absolute path,
+// leading separator dropped, remaining separators and drive colons flattened to '-', wrapped in
+// '--'. Mirrored here so Outfitter runs share pi's native per-project folder rather than starting a
+// parallel layout.
+const encodeProjectDirectory = (projectDirectory: string): string =>
+  `--${resolve(projectDirectory)
+    .replace(/^[/\\]/u, '')
+    .replace(/[/\\:]/gu, '-')}--`;
+
+/** pi's durable session folder for one project: `~/.pi/agent/sessions/--<encoded project>--`. */
+export const resolvePiUserSessionDirectory = (homeDirectory: string, projectDirectory: string): string =>
+  join(resolvePiUserAgentDirectory(homeDirectory), 'sessions', encodeProjectDirectory(projectDirectory));
+
+/**
+ * Session directory for a pi launch, or `undefined` when the launch should keep whatever
+ * `PI_CODING_AGENT_SESSION_DIR` it inherits — an explicitly configured store, such as a resident
+ * agent's workspace/PVC path, stays in control. `env`, `homeDirectory`, and `projectDirectory` are
+ * injected so the resolution stays pure and unit-testable.
+ */
+export const resolvePiSessionDirectory = (
+  env: Readonly<Record<string, string | undefined>>,
+  homeDirectory: string,
+  projectDirectory: string,
+): string | undefined => {
+  const inherited = env[PI_SESSION_DIRECTORY_ENV];
+  if (inherited !== undefined && inherited.trim() !== '') return undefined;
+
+  return resolvePiUserSessionDirectory(homeDirectory, projectDirectory);
+};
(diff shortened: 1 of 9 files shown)


ncrmro:
## Code review

**Verdict: approve.** The change is small, well-scoped, and the one subtle part — *which* durable directory — is the right choice for a non-obvious reason. Notes below are advisory; nothing blocks merge.

### What it does

`RunAgentCommand` resolves a durable pi session directory and threads it through `ProjectionInput.sessionDirectory`; `ProjectHarness` puts it in the pi launch env as `PI_CODING_AGENT_SESSION_DIR`. Non-pi harnesses and inherited values are left alone.

### Correctness

- **The per-project folder is load-bearing, not cosmetic.** In `pi-coding-agent` `core/session-manager.js`, an explicit session directory is used verbatim and flips `--continue`/`--resume` onto a cwd-filter path (`filterCwd = sessionDir !== undefined && dir !== getDefaultSessionDirPath(cwd)`), which reads every transcript header in the directory. Targeting `~/.pi/agent/sessions` would have flattened all projects into one folder *and* put resume on the slow path. Because `resolveProjectDirectory()` is `process.cwd()` — exactly the cwd pi inherits from `spawnLauncher` — the computed path equals `getDefaultSessionDirPath(cwd)`, so `filterCwd` stays `false` and pi keeps its fast path. Verified.
- **Precedence chain holds.** `spawnLauncher` merges `{...process.env, ...plan.env}` and `resolveAgentLaunchExecutable` spreads `...launchPlan.env` last, so the value survives to pi; pi then resolves `--session-dir` → env → `settings.json`, leaving pass-through flags in control.
- **Windows encoding matches pi.** `C:\Users\x` → `--C--Users-x--` under both pi's regex pair and the mirrored one here (the leading-separator strip is a no-op on a drive letter, then `:` and `\` both flatten to `-`).

### Risks worth naming

- **Coupling to a non-public pi detail.** `getDefaultSessionDirPath` is not in pi's `exports` map, so the encoding is mirrored rather than imported. If pi changes it, Outfitter silently starts writing to a parallel folder. It degrades gracefully — still durable, still project-scoped, and `filterCwd` would then correctly kick in — but prior history stops being found. The source is named in the comment, which is the right mitigation short of pi exporting the helper.
- **Sessions are project-scoped, not profile-scoped.** `outfitter run founder` followed by `outfitter run reviewer -- --continue` resumes the other profile's conversation under a different system prompt. This is deliberate (it is what "project-aware" in #223 means, and what a standalone `pi` does across differing `--system-prompt` invocations), but it is the one behavioral surprise in the change and is worth watching for user reports.
- **`state.md` tension.** The declared "Pi state paths" block still lists `sessions/: symlink`, describing the not-yet-implemented `state_persistence` engine rather than what ships. The new section plus the cross-reference comment covers it; a full reconciliation belongs with that engine, not here.

### Tests

Unit coverage hits the encoding, both opt-out shapes (blank vs. set env), pi-vs-claude projection, and a two-run `executeRunAgentCommand` case asserting transcripts outlive the deleted projection. New module is at 100%; suite is 302 passing, 99.09% statements against a 98% threshold.

The behavior was also exercised against the real bundled pi 0.80.3 under an isolated `HOME`: first run wrote to `<home>/.pi/agent/sessions/--<project>--/`, `--continue` resumed it and appended to the same file, `--no-session` recorded nothing, and an exported `PI_CODING_AGENT_SESSION_DIR` won and left the durable folder untouched.

### Nits (addressed in 64da931)

- The `resolvePiSessionDirectory` doc claimed the resolution is "pure"; `path.resolve` bases a relative `projectDirectory` on `process.cwd()`. Not reachable today, but the wording now states the expectation instead of overclaiming.
- Added the `state.md` cross-reference described above.

PR #248, opened 2026-08-03
fix(claude): append prompt documents through the flag Claude actually reads

ncrmro:
## The defect

The Claude adapter emitted one `--append-system-prompt <path>` per fragment. That flag takes a prompt **string** — handed a path it appends the path text, so the document contents never reach the model. Repeats are also last-wins, so even with the right form only the final fragment would survive.

An agent declaring `append_system_prompt` therefore launched on Claude with none of it applied, and nothing errored.

## Verified against the shipped binaries

Not inferred from the source:

```
claude --append-system-prompt      /tmp/a.md   -> path text, content lost
claude --append-system-prompt-file /tmp/a.md   -> content applied
claude --append-system-prompt-file a --append-system-prompt-file b -> only b
pi     --append-system-prompt a --append-system-prompt b           -> both
```

Reversing argument order reverses which survives on Claude, so the last-wins result is not a model-attention artifact.

## The fix

Claude now receives a single `--append-system-prompt-file` over a concatenation of the fragments in composition order — the native flag `OFTR-006.5.5` scopes prompt control to.

Pi is unchanged. Its `--append-system-prompt` reads a file path and accumulates across repeats, and it rejects the `-file` form outright (`Unknown option`). The two harnesses have no common spelling, so the divergence has to live in the adapter.

## New: `outfitter run --append-prompt <path>`

Repeatable, for documents the caller supplies at launch — a persona, typically. These land after the agent's own fragments, which is the position `OFTR-005.3.1` already reserved for *"runtime passthrough append prompts"*.

Passthrough after `--` cannot serve this: those arguments reach the harness verbatim, so a caller would have to know which harness will be launched and spell the flag its way — exactly the trap above. Every persona launcher and doc in the wider ecosystem currently passes a bare path this way, which works on pi and silently no-ops on Claude.

## A test asserted the bug

`run-agent-inheritance.test.ts`'s Claude ordering test expected repeated `--append-system-prompt` naming each fragment by path, and passed. That is why this survived in a shipped, tested feature.

Its requirement is unchanged — composition order is still asserted exactly, now observed inside the composed file. Its citation is corrected from `OFTR-006.3` (pi launch controls) to `OFTR-006.5` (Claude launch controls), and it carries a comment recording why the mechanism moved.

## Verification

- 308 tests pass; typecheck, eslint, prettier, coverage clean.
- 6 new tests: both harnesses, ordering, `--append-prompt` folding, trailing-newline separation, passthrough staying last.
- End-to-end with the built `cli.js` — Claude emits one `-file` flag containing agent body then persona in order; a real pi run returned the persona's codeword.

**Limits:** a live Claude run through `outfitter run` is auth-gated (the projection points `CLAUDE_CONFIG_DIR` at a throwaway root), so the Claude leg was verified at the launch-plan boundary rather than against a live model. The `paths.length === 0` guard is unreachable through composition and carries a `v8 ignore` explaining why it is kept.

### DIFF
--- a/code/cli/src/cli/commands/RunAgentCommand.ts
+++ b/code/cli/src/cli/commands/RunAgentCommand.ts
@@ -48,6 +48,8 @@ export interface RunAgentInput {
   readonly harness?: string;
   readonly strict?: boolean;
   readonly passThroughArgs?: readonly string[];
+  /** `--append-prompt` documents appended to the system prompt after the agent's own, in order. */
+  readonly appendPromptPaths?: readonly string[];
   readonly launcher: AgentProcessLauncher;
   /** Onboarding to run once when no agent is selected and settings define no `default_agent`. */
   readonly setup?: SetupRunner;
@@ -199,6 +201,7 @@ export const executeRunAgentCommand = async (input: RunAgentInput): Promise<RunA
       homeDirectory: input.homeDirectory,
       sessionDirectory: resolveSessionDirectory(input, harness),
       passThroughArgs: input.passThroughArgs,
+      appendPromptPaths: input.appendPromptPaths,
       extensionLoadDirs: harness === 'pi' ? extensions.loadDirs : undefined,
       // ProjectHarness only overlays these for the pi harness, so pass them through unconditionally.
       configurationOverlayDirectories: configurationOverlays,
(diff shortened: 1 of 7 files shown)


ncrmro:
## Adversarial review found the first commit fixed half the bug

`--system-prompt` has **the same defect** as `--append-system-prompt`. It also takes a prompt string, so the composed base prompt — the first element of the `OFTR-005.3.1` order — was arriving as a literal path and being discarded too. I fixed the append half and left the base half broken.

```
claude --system-prompt      /tmp/probe.md --print 'Say hello.'  -> persona ignored
claude --system-prompt-file /tmp/probe.md --print 'Say hello.'  -> persona applied
```

`--system-prompt-file` is **undocumented** — it appears in no `claude --help` output — but it applies the file's contents. Both prompt flags now go through one helper so the pair cannot drift apart again.

Two more findings, both real:

- **`--append-prompt` paths were never validated.** On Claude the first thing to touch the path was `readFileSync` inside projection, surfacing a raw `ENOENT` that never named the flag. On pi, Outfitter never opened the path at all, so the same typo failed differently per harness. Now rejected up front by `assertReadableAppendPrompts`, matching the existing `assertNoSettingsIssues` idiom.
- **Single-newline join merged documents.** A document ending mid-sentence ran into the next one's opening paragraph, and a following `#` heading became a setext heading for the preceding line. Now blank-line separated.

## Requirements amended

The reviewer correctly flagged that the test asserted `not.toContain('--append-system-prompt')` — the literal negation of the flag `OFTR-006.5.5` named. The requirement text described flags that do not work:

- **OFTR-006.5.6, .7** — the Claude adapter MUST use the `-file` forms and at most one append flag.
- **OFTR-005.1.9–.11** — `--append-prompt` is repeatable, applied after the composition's own fragments, validated before launch, and projected per harness because passthrough is unaltered by `OFTR-005.1.7`.

## Test gaps closed

Added traceability comments and a **Commander-path test**. The flag's entire user-facing surface — the repeat collector, the kebab-to-camel mapping, the interaction with `allowUnknownOption` and the variadic passthrough — was previously unexercised; only the internal `executeRunAgentCommand` was covered.

check-ci green: 311 tests, lint, prettier, coverage, docs. End-to-end against the built CLI now shows both `--system-prompt-file` and `--append-system-prompt-file` carrying real content, with no bare flag emitted.

PR #256, opened 2026-08-06
feat(projection): project loadout `tools` for pi and claude

### LINKED ISSUE #255: Project loadout `tools` for pi and claude — split from #183, with an availability-vs-permission mismatch
Splitting the **tools** bullet out of #183. The other four elements there are capability plumbing; `tools` is a security control, and it has a per-adapter semantic mismatch that the bundled framing hides. Verified against `origin/main` at v1.4.0.

## Today

`ProjectHarness.ts` never projects it, and says so:

```
// plugins/tools remain unsupported pending incremental parity (#183)
const supportedElements = (input: ProjectionInput): readonly string[] => {
  const baseline = ['skills', 'model', 'thinking'];
  if (input.harness !== 'pi') return baseline;
  const pi = [...baseline, 'subagents', 'mcp', 'prompt_template'];
  return input.extensionLoadDirs === undefined ? pi : [...pi, 'extensions'];
};
```

Everything upstream of that step works. `composeTools` merges `allow`/`deny` across the chain (`composer/Composer.ts:168-175`), it lands in `Loadout`, and `loadoutElementsInUse` reports it — so `unsupportedElements` returns it and `--strict` makes its presence fatal:

```
harness 'pi' cannot project loadout element 'tools'.
Strict mode: composition warnings and unsupported elements are fatal.
```

`tools:` is therefore accepted by the parser, merged by the composer, visible in `outfitter dump`, and fatal at launch. A profile that declares it fails closed at start-up rather than running unrestricted — which is the right failure, but it reads as a supported field right up until the run.

## The mapping already ships, for subagents

`projection/Materialize.ts` does exactly this projection in two places — `serializeSubagent` and `serializeComposedSubagent`:

```js
const denied = new Set(definition.loadout.tools?.deny ?? []);
const tools = definition.loadout.tools?.allow?.filter((tool) => !denied.has(tool));
```

and writes the result as a `tools` frontmatter list. The support matrix marks Subagents `Supported | Supported`. So `tools` is functional for delegates and fatal for the agent that delegates to them.

## Why it should not be bundled with the rest of #183

The two harnesses' native controls are not the same kind of thing:

| Harness | Flag | Semantics |
| --- | --- | --- |
| pi | `--no-builtin-tools --tools a,b,c` | **availability** — the tool does not exist in the session |
| claude | `--allowedTools a b c` | **permission** — the tool exists; the flag governs whether use needs approval |

A naive `allow → --allowedTools` projection gives one adapter a hard ceiling and the other a prompt-bypass list. Same profile, materially different posture, no error.

That matters more than the usual parity gap because this is the element people reach for to keep an agent away from a shell. **For a security control, "supported with silent precision loss on one adapter" is worse than "unsupported and fatal on one adapter"** — which is what pi does today, and is defensible. Whatever lands should not trade the second for the first.

Claude Code also has no `settings.json` materialization today: `buildLaunchPlan` gives it `CLAUDE_CONFIG_DIR` and nothing writes a `permissions` block into it (`permissions` does not appear anywhere in `code/cli/src`). `docs/architecture/controllable-elements.md` describes hooks as living in "the generated `settings.json`", but the hooks doc records that as a roadmap TODO. So a permissions-based projection would be new materialization, whereas `--allowedTools` / `--disallowedTools` are available on the CLI now.

## Open question

For a headless resident agent — `--mode rpc`, no human attached — does a Claude tool call that requires approval fail closed, or does it hang or auto-proceed? If it reliably denies, `--allowedTools` may be a faithful projection of pi's allowlist for this class of agent, and the mismatch above is confined to interactive use. I have not verified the non-interactive path and would not rely on it as a security boundary without that answer.

## Suggested shape

- Add `tools` to the **baseline** in `supportedElements`, not the pi branch.
- One `toolArgs` helper, branching per harness, reusing the allow/deny filter that already exists in `Materialize.ts`.
- Decide `allow` alone → does it imply `--no-builtin-tools`? (Otherwise `allow` means nothing on pi.)
- Decide `deny` alone → pi has no deny flag; filter the builtin list, or reject the combination?
- Document the availability-vs-permission distinction wherever `tools` is described, whichever way it resolves.

## Motivation

Running a resident agent that answers on GitHub issues, restricted to reading threads and posting comments. The intended profile field is unusable, so the restriction is applied instead by a custom entrypoint baked into an org-owned image:

```
--no-builtin-tools --tools channel_read,channel_respond,mcp
```

Three consequences:

1. The tool set becomes a property of the **image**, shared by every agent running it. A shell-capable dev agent cannot run alongside a locked-down responder without a second image.
2. It is pi-specific. Moving the same profile to Claude Code silently drops the restriction, because those flags are not Claude's.
3. Because a shell is unavailable, GitHub access has to go through an MCP server behind the permitted `mcp` tool — and `mcp` does not constrain which server tools are reachable. The server's own surface became the real boundary, and under broad toolsets that surface included repository writes. The tool restriction was doing less than it appeared to.

None of that needs a new concept — only the projection step. Happy to send a PR if the semantics above are settled first.

ncrmro:
Closes #255.

Projects the `tools` loadout element into native launch flags for both harnesses. The composer already merged `allow`/`deny` (`composeTools`); only the projection step was missing, so `tools:` parsed, merged, appeared in `outfitter dump`, and then killed the run under `--strict`.

## Behaviour

| Selection | pi | claude |
| --- | --- | --- |
| `{allow:[read,bash]}` | `--no-tools --tools read,bash` | `--tools read bash --allowedTools read bash` |
| `{deny:[bash,write]}` | `--exclude-tools bash,write` (no ceiling on the rest) | `--disallowedTools bash write` (no ceiling on the rest) |
| `{allow:[read,bash,write], deny:[bash]}` | `--no-tools --tools read,write --exclude-tools bash` | `--tools read write --allowedTools read write --disallowedTools bash` |
| `{allow:[bash], deny:[bash]}` | `--no-tools --exclude-tools bash` (zero-tool session) | `--tools "" --disallowedTools bash` (empties the built-in set; see caveat) |

`tools` absent → no args, nothing unsupported, both harnesses. Every combination is expressible, so `tools` is never reported unsupported.

The pi rows were **measured against pi 0.83.0** with a probe extension that registers a uniquely-named tool and prints `pi.getActiveTools()` at `session_start` (runs recorded in the #255 thread). The Claude rows come from `claude --help` and the CLI reference (code.claude.com/docs/en/cli-reference) — documented behaviour, not locally measured.

## Decisions this makes

#255 posed several of these as open questions. Answering them here rather than leaving the field unusable; each is reversible and argued in-code.

1. **An allowlist projects to TWO flags on Claude.** `--tools` is the availability ceiling (an unlisted builtin is not in the session) and `--allowedTools` pre-approves the same names within it (so a headless session is not stopped by a permission prompt for the very tools the profile granted). Either alone would be wrong: `--allowedTools` alone leaves every unlisted builtin available, merely unapproved; `--tools` alone leaves the granted tools prompting.
2. **pi's allowlist rides `--no-tools --tools a,b,c`, not `--no-builtin-tools`.** Measured: `--tools` is already a hard allowlist across built-in, extension, and custom tools; `--no-builtin-tools` keeps extension/custom tools by design, so it cannot express the ceiling. `--no-tools` is emitted alongside anyway — it costs nothing and keeps the allowlist a ceiling if a future pi made `--tools` additive.
3. **Deny-only is supported on both harnesses.** `--exclude-tools a,b` on pi, `--disallowedTools a b` on Claude; neither imposes a ceiling on the tools it does not name.
4. **The deny list is ALWAYS emitted, on both harnesses**, even when `allow` already removed those names. On Claude a bare name in `--disallowedTools` removes the tool from context per the docs, making the explicit denial an availability statement rather than a bet on the approval path; on pi it is belt and braces. OFTR-003.10.4 ("denied tools MUST win when projected") is satisfied without depending on the filter alone.
5. **An emptied allowlist is still a request, and both harnesses honour it.** pi: `--no-tools` (a true zero-tool session — built-in, extension, and custom tools all gone). Claude: `--tools ""`, the documented "disable all tools" form, passed as one empty-string argv element. Emitting only the denies here would fail open — every non-denied builtin would stay available though the profile requested zero.

## Open caveat: `--tools` scope on Claude

Per the CLI reference, Claude's `--tools` governs the **built-in set only**. MCP tools (`mcp__server__*`) are unaffected by it; they are governed by which MCP servers the loadout selects, and by `--disallowedTools` patterns such as `mcp__*` — a Claude-only pattern syntax a harness-neutral profile cannot lean on (`*` is a literal character to pi). So `--tools ""` is close to, but not exactly, pi's zero-tool session when MCP servers are selected: the builtins are gone, the MCP tools remain. This is documented in the `Tools.ts` block comment and both docs pages. Determined from `claude --help` and code.claude.com/docs/en/cli-reference; not verified by launching a session.

## Tool-name and shape validation

Projection puts these names directly into harness argv, so the same invariant is enforced at three depths, and a selection that cannot be projected exactly as written never launches:

1. **`agent.schema.json`** rejects a name starting with `-` or containing a comma or whitespace in `agent.md` frontmatter (`$defs/toolName`; a unit test pins the schema pattern to `TOOL_NAME_PATTERN`). A leading `-` would become an independent Claude flag (`allow: [Read, --dangerously-skip-permissions]` would turn off the permission system this element configures — a supply-chain path, since profiles compose from remote catalogs); a comma would split one name into several inside pi's `--tools` value.
2. **The agent reader** validates the RAW `tools` value of each `config.json` layer, which bypasses the schema — only `allow`/`deny` keys, each an array of non-empty strings — as a hard error naming the defect and the file. This exists because silent normalization fails open: `{"tools":{"allowed":["bash"]}}` or `{"tools":{"allow":"read"}}` previously collapsed to an empty selection and launched WITHOUT the intended restriction, `--strict` included. The reader also preserves key presence (an absent `allow` stays absent rather than becoming a ceiling-imposing empty list), and re-checks names on the merged record, since an overlay can replace `tools` wholesale after the schema check.
3. **`toolArgs` throws** on an unprojectable name even when `deny` would have filtered it, because `projectComposition` is exported and reachable without the read boundary.

## Also

- Extracts the allow/deny filter that was duplicated in `serializeSubagent` and `serializeComposedSubagent` into `effectiveToolAllowlist`; all three call sites share it.
- `supportedElements` now takes the composition; `tools` sits in the baseline for both harnesses.
- Tool args lead `launch.args`: Claude now has three variadic tool flags, emitted in the fixed order `--tools`, `--allowedTools`, `--disallowedTools` so each present flag is terminated by the next, with the prompt flags terminating the last. Placed later they would swallow a pass-through positional.
- Updates `docs/documentation/support-matrix.md` (Tool availability was `Roadmap | Roadmap`) and `docs/architecture/controllable-elements.md` to the corrected availability/permission story.
- Stale comment in `ProjectHarness.ts` corrected — `plugins` is still pending #183, `tools` no longer is.

## Verification

`tsc --noEmit`, `eslint`, `prettier --check`, and `npm run build` clean. 32 files / 341 tests pass; coverage 99.08% statements / 98.14% branches against a threshold of 98.

Two REQUIREMENT-marked tests changed with their requirements intact: the OFTR-003.10.4 assertion gains the `--tools` ceiling in the expected argv while still proving denied names are absent from every allow surface and `--disallowedTools` stays explicit; the OFTR-003.1/003.6 parse assertion drops a fabricated `deny: []` the author never wrote.


ncrmro:
Both review comments were correct and are fixed in `12fc95d` and `2d54a92`.

## Argument injection

Good catch, and it is the more serious of the two — it reproduced inside this PR the exact class of defect the PR was filed to argue against: a control that reads as a boundary and is not one.

Fixed at both layers, because they fail differently.

**Schema.** `agent.schema.json` gains `$defs/toolName`, referenced by `allow` *and* `deny` — `deny` is argv-spread on Claude too, so it is an equal vector:

```json
"pattern": "^[^\\s,-][^\\s,]*$"
```

Rejects a leading `-`, any comma, any whitespace. Mid-name hyphens stay legal (`web-search`).

**`toolArgs`.** Validates both lists and throws, naming the value. This is not redundant with the schema: schema rejection only covers definitions read through `readAgentDefinition`, and `projectComposition` is exported and reachable without it.

It **throws** rather than filtering, deliberately. A silently dropped entry in a security control is the failure mode this PR exists to reject.

Validation runs *before* deny-filtering, so `{allow: ['-x'], deny: ['-x']}` is still refused rather than quietly filtered clean.

`TOOL_NAME_PATTERN` is exported and a test asserts `schema.$defs.toolName.pattern === TOOL_NAME_PATTERN.source`, so the two layers cannot drift.

### Exposure, stated precisely

- **Claude: exploitable by construction.** Each entry becomes its own argv element, so a `--flag` in the list is parsed as an independent flag. Real escalation.
- **pi: no equivalent injection, and I am not claiming one.** Entries are comma-joined into a single argv element as the value of `--tools`, so no entry can become its own token. A comma inside a name splits it into more entries of the same allowlist the author already controls — self-widening, not escalation. One residual hazard in principle: if the first entry begins with `-` the joined value begins with `-`, and whether pi's parser treats that as the option value or promotes it to a flag is parser-dependent. Not verified; the validation closes it either way.
- **Whitespace is not a separator on either path.** `cross-spawn` is called without `shell: true`, so no shell parses these. Banning whitespace is conservatism, not a live vulnerability.

### One limitation worth flagging

The pattern also excludes Claude's scoped permission rules (`Bash(npm run test:*)`), which contain spaces. Deliberate for a harness-neutral field — that string is meaningless to pi — and the direction is chosen for compatibility: relaxing the pattern later is backward compatible, accepting it now and restricting later would not be.

## Stale matrix row

Fixed — `controllable-elements.md:210` is now `Partial | Supported`. The rest of that file's matrix was checked for other stale tools rows; there are none.

## Verification

typecheck, lint, prettier and build clean; 330 tests pass (5 added), coverage 99.11% statements against a 98 threshold.

CI has not run on this PR — no Actions run has been created in this repository since 2026-08-04T18:44Z, which appears to be an org-level halt rather than anything about this branch. All figures above are from a local run and still need CI confirmation before merge.

ncrmro:
All three correct, and the first two overturn a premise this PR was built on. Fixed in `32805e4` and `1e60945`.

I did not want to resolve this from the help text, which is ambiguous, so I measured it. pi has no tool-listing flag and RPC `get_state` does not report tools, but the extension API does — `pi.getActiveTools()`. A probe extension registering a uniquely-named `probe_marker_tool`, run as `pi --mode rpc --no-session --offline --no-extensions -e probe.ts <FLAGS> </dev/null` against pi 0.83.0:

```
(no flags)                                ["read","bash","edit","write","probe_marker_tool"]
--tools read                              ["read"]
--no-builtin-tools                        ["probe_marker_tool"]
--no-tools                                []
--no-builtin-tools --tools read           ["read"]
--no-tools --tools read                   ["read"]
--exclude-tools bash                      ["read","edit","write","probe_marker_tool"]
--exclude-tools bash,probe_marker_tool    ["read","edit","write"]
--tools read,bash --exclude-tools bash    ["read"]
--no-builtin-tools --tools read -xt read  []
```

The baseline row confirms the probe tool registers and is visible, so its absence elsewhere is a real exclusion rather than a registration failure.

## What the measurement corrects

**`--no-builtin-tools` was never required.** Row 2: `--tools read` alone yields `["read"]`. The claim in my earlier block comment, commit message, both docs pages and issue #255 — "`--no-builtin-tools` is REQUIRED whenever `allow` is set; without it `--tools` only adds to the builtin set" — is **false**. `--tools` is a hard global allowlist across built-in, extension and custom tools on its own. Rows 5 and 6 show the base flag makes no difference once `--tools` is present.

**`--no-builtin-tools` is not a zero-tool mode, and the consequence was worse than described.** Row 3 leaves the extension tool live. The emptied-allowlist branch was emitting a "zero tools" projection that still exposed every extension-registered tool. Now `--no-tools`, measured `[]`.

**Deny-only was projectable all along.** `--exclude-tools` works standalone (row 7) and composes with `--tools` (row 9). Leaving it unsupported meant `--strict` killed runs that had a perfectly good native form — and I argued for that at length as principled fail-closed design. The principle was fine; the fact under it was not.

## Resulting pi projection

| Selection | pi |
| --- | --- |
| allow-only | `--no-tools --tools read,bash` |
| deny-only | `--exclude-tools bash,write` (no ceiling on unnamed tools) |
| allow+deny | `--no-tools --tools read,write --exclude-tools bash` |
| empty-after-filter | `--no-tools --exclude-tools bash` |

Claude is unchanged. `--no-tools` is emitted alongside `--tools` even though rows 5/6 prove it redundant: it costs nothing, states the ceiling in the argv, and holds if a future pi made `--tools` additive. The block comment says explicitly that it is belt-and-braces rather than a requirement, so the old wrong claim cannot be re-derived from the code.

`toolsProjectable` always returned true once deny-only became projectable, so it is deleted and `tools` now sits in the plain baseline. It is never unsupported, on either harness.

## `config.json` bypass

Confirmed as described: `parseAgentDefinition` schema-validates frontmatter and *then* merges overlays, and `loadoutFromRecord` accepted every string. The merged loadout is now re-checked against the same `invalidToolName` predicate `Tools.ts` uses, so the layers cannot diverge. Four tests in a new `agent-definition-tools.test.ts` cover overlay `allow`, overlay `deny`, the frontmatter path, and a positive case.

## Caveat

Every measurement above is pi **0.83.0**. This repo pins `@earendil-works/pi-coding-agent ^0.80.3`; I did not measure a 0.80.x install, and the flag semantics could differ.

## Verification

typecheck, lint, prettier and build clean; 334 tests pass across 32 files; coverage 99.11% statements against a 98 threshold. `Tools.ts`, `ProjectHarness.ts` and `AgentDefinition.ts` at 100%.

CI still has not run — no Actions run has been created in this repository since 2026-08-04T18:44Z, which looks like an org-level halt rather than anything about this branch. Every figure here is from a local run.

PR #258, opened 2026-08-07
fix(extensions): compare the pinned git ref before trusting a cached extension

ncrmro:
The pi-extensions git cache is revision-blind: `ensureOneExtension` checks only `existsSync(installDir)`, so an existing cached checkout wins regardless of the profile's `@ref` pin. Bumping a pinned extension revision in a profile silently no-ops — observed in production, where a downstream deployment carries a shell script that parses the pin out of `agent.md` and `rm -rf`s the cache directory whenever it moves. This PR makes the cache verify the pin and obsoletes that workaround.

## Behavior

- **40-hex SHA pins**: the cached checkout's `git rev-parse HEAD` is compared to the pin — offline-safe, no network. Mismatch or unreadable HEAD → reinstall.
- **Branch/tag refs**: a sibling marker `<installDir>.outfitter-ref.json` records `{ref, headSha}` at install time; reinstall only when the requested ref *string* changes. A moved remote branch tip is deliberately not chased — pinned launches never hit the network.
- **Legacy caches without a marker** (non-SHA): served as-is offline (dropping them would break air-gapped deployments on upgrade), refreshed once when online, which writes the marker.
- **Offline + provably stale**: warn-and-drop, naming the extension, the pin, and what was found — mirroring exactly how the offline path treats a missing extension (warning; fatal only under `--strict`).
- **Reinstall is remove-then-install**: `pi install` owns the directory layout, so a mid-way failure leaves neither dir nor marker and reads as plain-missing next run.
- npm specifiers byte-for-byte unchanged.

## Verification

11 new tests (SHA hit/stale, offline+stale warning, marker lifecycle, ref-change, legacy paths, garbage marker). Full suite 328/328, coverage 99.12% statements against 98 thresholds; tsc/eslint/prettier/build clean.

Unblocks the downstream cleanup tracked in ai-outfitter/outfitter#234's consumer story.

### DIFF
--- a/code/cli/src/extensions/PiExtensionCache.ts
+++ b/code/cli/src/extensions/PiExtensionCache.ts
@@ -17,7 +17,7 @@
 // Offline, a checkout that provably mismatches its pin is dropped with a warning — the same
 // severity the offline path already applies to a missing extension (fatal only under `--strict`).
 import { existsSync, mkdirSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
-import { join } from 'node:path';
+import { join, resolve, sep } from 'node:path';
 
 import { launchThroughSpawn, spawnLauncher } from '../agents/AgentLaunch.js';
 import { runGit } from '../sources/GitRepository.js';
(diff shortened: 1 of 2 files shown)


ncrmro:
The traversal finding was correct — and this PR made it worse than the reviewer stated: pre-change a traversal specifier was a mis-resolved cache hit; the new stale-reinstall path turned it into an `rmSync` of a directory outside the cache root. Fixed in 47671ff at both layers: every git path segment and npm name part is rejected if `.`, `..`, or backslash-bearing (before `installSegments` is built), and an exported containment assertion resolves the install dir and requires it strictly inside the cache root before any filesystem access — which also keeps the derived marker sibling inside. Segment rejection surfaces through the existing unsupported channel (warn, fatal under `--strict`); the containment backstop is a hard throw. New tests include a live proof: an outside directory survives `ensurePiExtensions(['git:../../victim@v1'])` with zero spawns. 332/332, coverage above the 98 gates.

Issue #212, opened 2026-07-23
Define remote catalog trust and provenance policy


ncrmro:
## Why

Remote catalogs can supply instructions, skills, scripts, extensions, and other executable or behavior-changing resources. Organization adoption needs a reviewable trust model: which sources are allowed, exactly what revision was resolved, and what provenance survives into validation, dumps, and future baked artifacts.

Planning material in #109 and #194 explored governance, but mixed source trust with broader enterprise/product claims. This issue narrows the work to portable source trust and provenance.

## Scope

- Source identity, normalized URI handling, and immutable revision pinning.
- Optional organization/project allowlists and explicit trust decisions.
- Provenance records for resolved resources, deterministic dumps, and future artifacts.
- Credential redaction and secret-safe audit output.
- Update/rotation behavior when a pin, source, or allowlist changes.
- A clear separation between implemented OSS behavior and any future enterprise enforcement.

## Plan

1. Threat-model remote catalog resolution and transitive resources.
2. Specify the provenance schema and its stable identifiers.
3. Define validation findings for unpinned, disallowed, unavailable, or changed sources.
4. Integrate with remote sync (#184), dumps, and artifact work (#187).
5. Add end-to-end fixtures for allowed, denied, moved, redacted, and offline sources.

## Acceptance criteria

- [ ] Users can identify the source and immutable revision of every resolved remote resource.
- [ ] Validation can enforce a documented source policy without exposing credentials.
- [ ] Dumps and future artifacts carry reviewable provenance.
- [ ] Offline and last-good-cache behavior cannot disguise a source or revision change.
- [ ] Documentation does not claim enforcement that is not implemented.

Related: #109, #194, #184, #187, #197.

PR #262, opened 2026-08-08
feat: resolve catalog-declared sources transitively (OFTR-004.6)

tylerwillis:
## What & why

A catalog can declare its own dependencies via `sources` in its settings file, but resolution and sync ignored them. The concrete failure: a fresh install of the pinned default catalog warns that the founder profile's `persona-authoring` and `persona-review` skills are unknown — they live in `community-profiles`, a source `default-profiles` declares (its 1.1.0 changelog: "pin community-profiles as a catalog source") but nothing ever fetched. The `sources` field looked load-bearing yet did nothing.

This makes the existing declaration mean what it says: **resolution and `outfitter sync` now follow a cached catalog's declared dependencies to a fixpoint, and setup fetches the pinned closure before launching a profile** — so a fresh `outfitter` install works with no manual sync.

Verified end-to-end against real GitHub: first-run bootstrap fetches the full `default-profiles` → `community-profiles` closure, and `outfitter sync` on a fresh home exits 0 with the founder profile's persona skills resolving from `community-profiles`.

**Per-source sync validation** also had to change: `sync` validates each fetched source in isolation, and the real `default-profiles` catalog references skills that live in its declared dependency — so isolated validation rejected it as "invalid" and nothing cached. Sync now **defers** an unresolved loadout slug to a warning (a source may reference a dependency's skill); loadout wholeness stays authoritatively enforced by `outfitter validate` (an error there), and the run-time composer keeps its non-fatal warning per OFTR-005.3.4.

## Safe subset (deliberate scope limit)

Transitive sources are restricted to the narrowest safe subset while the remote-catalog trust/provenance model (#212) is defined:

- **`github:` shorthand only.** A declared `uri:` is skipped with a warning. A URI could name an arbitrary git transport — an `ext::<cmd>` remote helper is code execution, a filesystem path escapes the sandbox — and a GitHub-spelled URI would bypass the private-catalog gate. Keeping to `github:` (schema-constrained to `owner/repo`, always normalized to `https://github.com/…`) routes every transitive fetch through the same gate as your own sources.
- **Whole repository only.** A transitive source may not carry a `path:` subpath, so a declaration can never point outside its fetched checkout.
- **Pinned to an immutable ref** (full SHA, or a version tag — a movable pin; SHA for a closure that never changes underneath you).
- **Content only** — no policy settings (default agent/harness, cache dir, `remote_settings`, enterprise) take effect transitively.
- **Direct sources outrank transitive ones**, breadth-first by depth; cycles resolve once.

The real use case (`default-profiles` → `community-profiles`, both `github:`) works; broader transports are future work under #212.

## Reviewed

Seven rounds of independent adversarial review (a GPT-family model + a fresh-context Claude agent each round, told to *break* the change and, later, to judge design quality). Findings converged round over round — Critical/High → Medium → Low — and each fix landed with a mutation-sensitive test:

- **Security surface** (safe-subset design): an `ext::` remote-helper RCE, a private-catalog gate bypass, a `path:` sandbox escape, and a settings-symlink DoS — all closed by restricting transitive sources to `github:` shorthand + immutable ref, anchoring settings containment to the checkout root, and guarding every payload-resolution throw.
- **Correctness**: a source-identity conflation (two-key model: path-less redacted key for the shared checkout cache, path-aware raw-URI key for graph/dedup — the latter also distinguishes credentials so a same-repo different-credential source isn't dropped); duplicate-source dedup across resolve/sync/bootstrap; a dangling-symlink `settings.yml` now warns rather than silently drops.
- **Coherence with the rest of the CLI**: `outfitter sync` validates each source in isolation, so the real default catalog — whose founder profile references a dependency's skills — was rejected; sync now defers unresolved loadout slugs (authoritatively enforced by `outfitter validate`, warned at runtime per OFTR-005.3.4). First-run onboarding surfaces a best-effort-bootstrap dependency failure as actionable `outfitter sync` guidance to both terminal and programmatic callers.

Verified end-to-end against real GitHub: a fresh install bootstraps the full closure; `outfitter sync` exits 0; `outfitter dump founder` composes with the persona skills materialized in its closure. One known-accepted item: a settings-read TOCTOU requiring concurrent `outfitter sync` of one cache — not a supported mode.

## Spec & tests

New requirements `OFTR-004.6` (transitive catalog sources), each MUST covered by a test with the two-line traceability comment. Full gate green locally: `typecheck`, `lint`, coverage (99%+, thresholds hold), docs build. Sync closure and bootstrap closure are tested hermetically via an injected repository-sync fake that maps `github:` fixtures to local repos.

## Reviewing this: core vs. hardening (the second is a candidate for #212)

This is one commit, but it's two separable concerns. If the hardening feels like it front-runs the trust model you want to design holistically in #212, it can be pulled into a follow-up (or folded into #212) without touching the core — happy to split it into stacked commits/PRs on request.

**Core — the fix (safe by construction; needed for the feature to work):**
- Follow a catalog's declared `github:` sources transitively, restricted to the safe subset (github shorthand + immutable ref, no `uri:`/`path:`): `sources/TransitiveSources.ts`, `resolver/Layer.ts`, sync fixpoint in `cli/commands/SyncCommand.ts`, bootstrap closure in `setup/DefaultCatalog.ts`.
- Sync loadout deferral (without it the real default catalog can't sync — it references a dependency's skills): `resolver/ResolverValidation.ts`, `SyncCommand.ts`.
- Two-key identity + duplicate-source dedup: `sources/SourceCache.ts`.
- Post-onboarding warning surfacing: `cli/commands/RunAgentCommand.ts`, `resolver/ResolverContext.ts`.
- Requirements `OFTR-004.6.1–.7, .10, .11`.

**Hardening — defense-in-depth against a hostile catalog's *contents* (candidate for #212):**
- Settings-file containment: checkout-root anchoring, dangling-symlink detection (`lstat`), precise `..` idiom, the accepted-TOCTOU note — `TransitiveSources.ts`.
- Path-escape crash guards (`resolveSourcePayloadRoot` try/catch) — `TransitiveSources.ts`, `Layer.ts`, `SyncCommand.ts`.
- Directory-not-file payload validation — `DefaultCatalog.ts`.
- Credential-distinguishing identity + path canonicalization — `SourceCache.ts`.
- Requirement `OFTR-004.6.8` (unreadable settings), part of `.9`.

**Why the line is drawn there:** `github:`-only is the *transport floor* and stays in the core — it keeps all three catalog types (the default catalog, community-profiles, and user-published github repos) and their transitive deps on a known transport, so it can't be separated out safely. The hardening above is baseline robustness for multi-catalog resolution; the *policy* layer — allowlists, provenance, "should I trust this transitive dep at all" — is #212's, and nothing here pre-empts those decisions.

## Related

Groundwork for the trust/provenance model tracked in #212.

### DIFF
--- a/code/cli/src/cli/commands/RunAgentCommand.ts
+++ b/code/cli/src/cli/commands/RunAgentCommand.ts
@@ -171,7 +171,6 @@ export const executeRunAgentCommand = async (input: RunAgentInput): Promise<RunA
   let resolved = resolveEffectiveSet(input);
   assertNoSettingsIssues(resolved.settingsIssues);
   assertReadableAppendPrompts(input.appendPromptPaths);
-  emit(resolved.warnings.map((warning) => `warning: ${warning}`));
 
   // First run: nothing selected and no default configured — onboard, then resolve again.
   const setupMessages: string[] = [];
@@ -188,13 +187,19 @@ export const executeRunAgentCommand = async (input: RunAgentInput): Promise<RunA
     }
   }
 
+  // Warnings from the FINAL resolved state (after any onboarding), folded into every returned message
+  // array after the setup notices — so a dependency that best-effort bootstrap could not fetch
+  // surfaces its actionable `outfitter sync` guidance to both the terminal and programmatic callers,
+  // not just a generic unknown-skill warning at composition (OFTR-004.6.10).
+  const resolutionWarnings = resolved.warnings.map((warning) => `warning: ${warning}`);
+
   const { set, settings } = resolved;
   const agentSlug = resolveAgentSlug(settings.defaultAgent, input.agent);
   const harness = resolveHarness(settings.defaultHarness, input.harness);
   const composed = compose(set, agentSlug, { projectDirectory: input.projectDirectory });
 
   if (composed.plan === undefined) {
-    const messages = [...setupMessages, ...composed.warnings, ...composed.errors];
+    const messages = [...setupMessages, ...resolutionWarnings, ...composed.warnings, ...composed.errors];
     emit(messages);
     return { exitCode: 1, messages };
   }
@@ -231,6 +236,7 @@ export const executeRunAgentCommand = async (input: RunAgentInput): Promise<RunA
     if (input.strict === true && warnings.length > 0) {
       const messages = [
         ...setupMessages,
+        ...resolutionWarnings,
         ...warnings,
         'Strict mode: composition warnings and unsupported elements are fatal.',
       ];
@@ -238,8 +244,9 @@ export const executeRunAgentCommand = async (input: RunAgentInput): Promise<RunA
       return { exitCode: 1, messages };
     }
 
-    // Emit setup notices and warnings before launch so they precede the pi session on the terminal.
-    const messages = [...setupMessages, ...warnings];
+    // Emit setup notices, resolution warnings, and composition warnings before launch so they precede
+    // the pi session on the terminal (and are returned to programmatic callers).
+    const messages = [...setupMessages, ...resolutionWarnings, ...warnings];
     emit(messages);
 
     // Attach the Outfitter runtime UI and sign-in extension to interactive pi sessions.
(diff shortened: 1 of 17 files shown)


tylerwillis:
@ncrmro this fixes a bug I hit testing fresh new user install (community profile store didn't load which broke some of the config for founder profile). 

I kicked it off before bed and it went aggressive in the reviews, apologies if it's over-engineered :)

PR #269, opened 2026-08-09
feat: add invocation-only runtime layers

tylerwillis:
## Why

Panopticon generates workflow and operation skills for one supervised agent run. Today it must either write those resources into the checkout or persistent user configuration, or translate them into Pi-, Claude-, or Codex-specific arguments itself.

The downstream flow is tracked in [Panopticon #353](https://github.com/ai-outfitter/panopticon/issues/353): [`container/agent.py`](https://github.com/ai-outfitter/panopticon/blob/main/src/panopticon/container/agent.py) prepares per-run workflow resources, then [`runner.py`](https://github.com/ai-outfitter/panopticon/blob/main/src/panopticon/runner.py) launches the selected harness. Outfitter should remain responsible for resolving those resources and projecting them through its existing adapters.

## What changes

```console
outfitter run engineer --runtime-layer /run/panopticon/resources
```

`--runtime-layer` accepts the contents of a protocol `.agents` directory for this invocation only. It is repeatable: runtime layers resolve above the workspace in command-line order, then the existing resolver and adapters behave unchanged.

The supplied roots and persistent settings are never modified.

## Deliberate boundaries

- This affects `run` only. `list`, `validate`, and `dump` do not have a demonstrated invocation-layer consumer.
- It does not change signal handling or process supervision; current main already owns that behavior through [#235](https://github.com/ai-outfitter/outfitter/pull/235).
- It adds no harness-specific projection. Pi, Claude, and Codex continue through the existing composition and adapter path.
- Generated overview text can use the existing `--append-prompt`; runtime layers are for protocol resources such as skills.

The focused contract test checks the three important properties: earlier runtime layers win, the winning skill reaches the projected runtime, and neither the input layer nor persistent settings are changed. The repository's full `check-ci` suite passes locally.

For review, the design questions are limited to whether a protocol resource root is the right input, whether invocation-above-workspace precedence is correct, and whether the non-persistence boundary is clear.

### DIFF
--- a/code/cli/src/cli/commands/RunAgentCommand.ts
+++ b/code/cli/src/cli/commands/RunAgentCommand.ts
@@ -48,6 +48,8 @@ export interface RunAgentInput {
   readonly harness?: string;
   readonly strict?: boolean;
   readonly passThroughArgs?: readonly string[];
+  /** Invocation-only protocol resource roots, highest precedence first. */
+  readonly runtimeLayers?: readonly string[];
   /** `--append-prompt` documents appended to the system prompt after the agent's own, in order. */
   readonly appendPromptPaths?: readonly string[];
   readonly launcher: AgentProcessLauncher;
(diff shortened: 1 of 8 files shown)


ncrmro:

PR #313, opened 2026-08-19
feat(claude): inherit native user configuration

### LINKED ISSUE #312: feat(claude): inherit the user's Claude configuration by default
### What part of Outfitter does this touch?

Harness adapter (pi, Claude Code, Codex, other)

### What can you not do today?

Someone who types `outfitter run <profile> --harness claude` expects to get *their* Claude wearing a
costume — same login, same trusted repos, same permission allowlist, same MCP servers and plugins,
same model and effort preferences — plus whatever the profile adds. What they get is a stranger's
Claude: a blank configuration directory that happens to share their credentials.

The reproducer: launching a profile in a workspace opens the workspace-trust modal, even though
native Claude never asks there.

`buildPiOrClaudeLaunchPlan` (`code/cli/src/projection/ProjectHarness.ts:163-171`) points
`CLAUDE_CONFIG_DIR` at a per-run `mkdtemp` directory that is deleted on exit. Exactly three values
are seeded into it (`code/cli/src/agents/ClaudeStatePersistence.ts:244-273`): `.credentials.json`,
`oauthAccount`, `hasCompletedOnboarding`, plus `projects[cwd].hasTrustDialogAccepted` when already
`true`. Every launch also passes `--strict-mcp-config` unconditionally.

What that costs, roughly in order of pain:

| Lost | Why |
| --- | --- |
| Workspace trust | the seed copies only the exact-path entry, so ancestor grants never arrive |
| `permissions.allow`/`deny`, `defaultMode` | `~/.claude/settings.json` is never read by any code path; a Claude projection has no `settings.json` at all |
| In-session permission approvals | written into the tmpdir; only `oauthAccount` is merged back, so the same commands are re-approved every run, forever |
| User and project MCP servers | suppressed by `--strict-mcp-config` |
| Plugins (`enabledPlugins`) | `~/.claude/plugins` is not projected |
| Hooks, statusline, model, effort, notification prefs | all live in the unread `settings.json` |

The trust case concretely: `~/.claude.json` records the workspace as `hasTrustDialogAccepted: false`
while its ancestor is `true`. Native Claude resolves trust through the ancestor chain; the seed
copies only the exact-path entry. The fix is to stop hand-picking keys, not to reimplement the
ancestor walk.

### What should it do?

Inherit the user's Claude configuration by default, with `outfitter run --isolated` (and an
`isolation: inherit | isolated` setting) as the opt-out for CI, containers, and profile authors
verifying reproducibility.

The principle: isolation was never the goal, precedence was. The profile should win **where it
speaks** — system prompt, skills, subagents, declared MCP servers, model, tool restrictions — and
the user's configuration should apply **everywhere the profile is silent**. Today's implementation
confuses "the profile wins" with "nothing else exists."

Secondary but design-constraining: a profile must still be able to *narrow* the inherited
permission set — share credentials and trust while running a review persona that cannot write.
Inheritance is the default, not a ceiling removal. That is why the settings we hand Claude has to be
a generated overlay file rather than a raw pointer at `~/.claude/settings.json`.

Two mechanisms are viable and a short spike picks between them; both are detailed in the plan
comment below.

### Related requirement

OFTR-006.5 (items 1, 11, 12, 13, 16 — this change contradicts them and must amend them first)

### Harnesses affected

Claude Code only; pi and Codex untouched.

### What are you doing instead?

Re-accepting the trust dialog and re-approving the same permission prompts on every single run, and
losing user-scope MCP servers and plugins for the duration of the session.

luce-unsup:
## Summary
- inherit Claude's native user settings, trust, MCP servers, plugins, and session history by default
- project composed skills and subagents through a temporary Claude plugin while keeping MCP projection additive
- retain the previous hermetic behavior behind `--isolated` and user-home `claude.isolation`, with compatibility fallback and inspectable startup reporting
- document both modes and update adapter requirements

## Validation
- `npm run check-ci`
- `npm run build`

Closes #312


ncrmro:
Closing in favor of #317, which implements the same design. This is not a rejection of the approach
— the two arrived independently at the same mechanism (`--plugin-dir` over the native configuration,
conditional `--strict-mcp-config`, `--isolated` opt-out, home-scope gating, a `claude --help`
compatibility probe), and this PR got two things right that #317 initially did not: the plugin
belongs in a `plugin/` subdirectory rather than at the projection root, and `docs/architecture/file_structure.md`
needed updating. Both are carried over.

The deciding issue is requirements traceability. This PR replaces `OFTR-006.5` items 12–19 with new
obligations at the same numbers, and updates no traceability comments. Those eight IDs are cited by
roughly twenty tests in `claude-state-persistence.test.ts`, so after a merge every one of them points
at the wrong obligation — `.12` (seed `.credentials.json` and the narrow `.claude.json` keys) would
mean "project selected skills and subagents as a temporary Claude plugin"; `.17` (recursively seed
the cwd's session history) would mean "profile controls remain explicit launch arguments". The
isolated-mode contract this PR states it preserves would also have no normative text left anywhere
in the repo, and `INTEGRATION_TEST_FIXTURES.md` line 60, which justifies unit coverage "for
OFTR-006.5 items 12–19", becomes false without being touched.

#317 keeps 12–19 intact, adds a scoping paragraph saying they govern the isolated strategy, and
appends 20–23 for the inherited one.

Two smaller behavioral notes, both verified against the diff rather than inferred:

- The compatibility probe defaults to off for direct callers: `input.probeClaudeInheritance?.() ?? []`
  means anything invoking `executeRunAgentCommand` as a library — which is what every test does —
  skips the check and assumes inheritance is safe. The `spawnSync` also has no timeout.
- `--strict` runs fail when their only problem is an old Claude, because the fallback warning is
  pushed into the `warnings` array that strict mode treats as fatal. A CI box with a Claude lacking
  `--plugin-dir` turns a should-be-quiet fallback into a hard failure.

Thanks — the design work here stands, and the comparison made #317 better.

ncrmro:
Correction to the comment above: only one of the two is carried over. The `file_structure.md` entry
for the new module is in #317. The `plugin/` subdirectory is deliberately **not** — isolated mode
still uses the projection root as `CLAUDE_CONFIG_DIR` and finds `skills/` and `agents/` there, so
moving them under `plugin/` would introduce a mode-dependent materialization path for a tidiness
gain with no observed behavioral difference. Worth revisiting if the isolated path is ever retired.

PR #316, opened 2026-08-20
fix(validate): diagnose missing Pi MCP adapter

### LINKED ISSUE #315: outfitter validate: warn when mcp: is declared with no MCP-capable extension
## Problem

A profile can declare `mcp:` entries referencing servers in its resolved `mcp.json` without declaring any extension that actually projects MCP into the session (e.g. `npm:pi-mcp-adapter` for `--harness pi`). Today that combination is silent: `outfitter validate` passes, `outfitter run` starts cleanly, and the agent has zero MCP tools — no error, no warning, anywhere.

Found live on 2026-08-19 (ai-outfitter/outfitter#314, closed as a test error once the missing extension was identified): a resident profile's `mcp:` entries went unprojected for the entire day because `extensions:` never listed `npm:pi-mcp-adapter`. The failure mode reads exactly like a harness bug — the agent runs, looks healthy, and simply has no MCP tools — and took a from-scratch pod-level repro to diagnose because nothing in `validate` or `run` pointed at the actual cause.

## Proposal

`outfitter validate` (and ideally `outfitter validate --strict`) should flag a profile whose composed `mcp:` list is non-empty but whose composed `extensions:` contains no MCP-capable adapter for the selected/target harness. A `harness`-aware check is fine — the requirement is harness-specific (`pi` needs `pi-mcp-adapter`; other harnesses may not need anything extra).

## Acceptance criteria

- `outfitter validate` against a fixture profile that declares `mcp: [<server>]` in a resolved `mcp.json` but no MCP-capable extension in `extensions:`, targeting `--harness pi`, prints a warning (or fails under `--strict`) naming the missing extension.
- The same fixture with `npm:pi-mcp-adapter` added to `extensions:` validates clean.
- A profile with no `mcp:` entries at all is unaffected (no new warning).
- Document the check in whatever reference `outfitter validate`'s existing checks are documented in.

Open a pull request referencing this issue when done.

https://claude.ai/code/session_01JSKm3EdfE3V8i35Rt6rC8B

luce-unsup:
## Summary
- add a harness-aware validation check for composed Pi MCP selections
- warn with actionable `npm:pi-mcp-adapter` guidance and preserve strict-mode failure semantics
- document the validation behavior and cover adapter, empty-MCP, non-Pi, and CLI-option cases

## Validation
- `npm run check-ci`
- `npm run build`

Fixes #315

### DIFF
--- a/code/cli/src/cli/commands/ValidateCommand.ts
+++ b/code/cli/src/cli/commands/ValidateCommand.ts
@@ -1,11 +1,15 @@
 // Provides `outfitter validate [--strict] [--json]` over the effective resource set.
 
-import { Command } from 'commander';
+import { Command, Option } from 'commander';
 
+import { compose } from '../../composer/Composer.js';
+import { listResources } from '../../resolver/Resource.js';
 import { resolveEffectiveSet } from '../../resolver/ResolverContext.js';
 import type { ValidationFinding } from '../../resolver/ResolverValidation.js';
 import { validateEffectiveSet } from '../../resolver/ResolverValidation.js';
 import { formatSettingsIssue } from '../../settings/SettingsLoader.js';
+import type { Harness } from '../../settings/Settings.js';
+import { HARNESSES } from '../../settings/Settings.js';
 import type { CommandObject } from './CommandObject.js';
 import { resolveHomeDirectory, resolveProjectDirectory } from './ProcessDefaults.js';
 
@@ -14,6 +18,7 @@ export interface ValidateInput {
   readonly projectDirectory: string;
   readonly strict?: boolean;
   readonly json?: boolean;
+  readonly harness?: Harness;
 }
 
 export interface ValidateResult {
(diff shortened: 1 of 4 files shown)


ncrmro (changes requested):
Reviewed by an independent Opus pass, checked out in an isolated worktree, tests re-run against the real code plus mutation testing.

**Verdict: the feature is correct — all four acceptance criteria from #315 are genuinely met — but one test is vacuous and should be fixed before merge.**

## Confirmed defect

`code/cli/tests/unit/validate-mcp-extension.test.ts:71-82` — "wires the `--harness` target through the validate command" doesn't discriminate. It parses `--harness pi`, but `pi` is also the fallback default (`ValidateCommand.ts:60`). Mutation-tested: deleting `harness: options.harness` from the action handler at `ValidateCommand.ts:108` still passes all 4 tests unchanged. The test cannot fail for the thing it claims to check.

This is worth fixing rather than waiving because the file carries a `// THIS TEST VALIDATES A HARD REQUIREMENT ... YOU MUST NOT MODIFY` banner — merging as-is enshrines a non-discriminating test behind a guarantee it doesn't provide. One-line fix: also parse `--harness claude` against the same fixture and assert the adapter warning is absent, as a genuine negative control (mirroring what tests 1/3 already do correctly for the harness-detection case).

## Verified clean (no action needed)

- Detection logic (`plan.loadout.mcp.length === 0` skip, `harness === 'pi'` gate) — both real, both correctly tested elsewhere in the file.
- `--strict` semantics — emitted as `severity: 'warning'`, flows through the existing `ok = !hasErrors && !(strict && hasWarnings)` gate, matching how shadowing/reserved-namespace findings behave.
- Harness resolution (`input.harness ?? settings.defaultHarness ?? 'pi'`) mirrors `RunAgentCommand.ts:109-117` exactly; probed against `.agents/settings.yml` — `default_harness: claude` correctly suppresses the check.
- Inheritance: an agent inheriting `npm:pi-mcp-adapter` from a parent does not false-positive.
- Docs (`cli.md:78`, OFTR-003.7) accurately describe the behavior and precedence.
- No unrelated changes; single commit on `main`.

## Two non-blocking notes worth knowing, not requesting either as a change

1. `isPiMcpAdapter` (`ValidateCommand.ts:40`) matches only `npm:pi-mcp-adapter[@ver]`; `PiExtensionCache.mapSpecifierToPiSource` also accepts `git:` sources, so a profile vendoring the adapter via `git:` would false-positive (and hard-fail under `--strict`). No in-repo profile does this today.
2. Defaulting to `pi` when `default_harness` is unset is a behavior change for claude-only CI users who never set it — consistent with `outfitter run`'s identical fallback, so it reads as deliberate, but worth knowing if anyone's `--strict` CI starts failing after this lands.

Requesting the one test fix; happy to re-review once it's in.

luce-unsup:
Addressed the vacuous CLI wiring test in 9e42922. It now parses both `--harness pi` and `--harness claude` against the same MCP fixture and asserts the warning is present only for Pi, so removing the action-handler forwarding makes the negative control fail. Revalidated with `npm run check-ci` (558 tests) and `npm run build`.

PR #320, opened 2026-08-20
feat(profiles): project trusted launch environment

### LINKED ISSUE #319: feat: let a profile declare its launch environment
### What part of Outfitter does this touch?

Profile composition or resolution

### What can you not do today?

A profile cannot carry the environment its model needs, so anything that is configured by
environment variable has to be exported by hand at every launch.

The case that surfaced it: running a profile against a local model. Ollama 0.32.4 serves an
Anthropic-compatible `/v1/messages`, so Claude Code drives it directly — but only if the process
starts with the right environment:

```
ANTHROPIC_BASE_URL=http://localhost:11434 \
ANTHROPIC_AUTH_TOKEN=ollama \
outfitter run luce --harness claude --model qwen3:4b
```

That works today (verified on Claude Code 2.1.226 / Outfitter 1.10.0), which is the point: the
launch is one `env:` block away from being a property of the profile rather than a property of the
shell that happened to invoke it. A "local model" profile is not portable, cannot be shared through
a catalog, and cannot be described by `outfitter dump` — the part that makes it work lives outside
Outfitter entirely.

The same gap applies to a proxy (LiteLLM, Bedrock/Vertex via their own variables), to
`CLAUDE_CODE_*` toggles a profile wants on, and to pi's provider variables.

### What should it do?

An agent definition should declare environment variables, and the adapters should project them into
the launch environment they already build:

```yaml
---
name: local
model: qwen3:4b
env:
  ANTHROPIC_BASE_URL: http://localhost:11434
  ANTHROPIC_AUTH_TOKEN: ollama
---
```

Notes for whoever picks this up:

- **The requirement already exists and is unimplemented.** `OFTR-006.5.3` ("The Claude Code adapter
  MUST support profile-controlled environment variables") and `OFTR-006.3.2` for pi. There is no
  `env` key in `code/cli/src/schemas/agent.schema.json`, nothing named `environment` in
  `AgentDefinition.ts` or `Composition.ts`, and `buildPiOrClaudeLaunchPlan` builds `env` from the
  harness alone. So this is closing a gap between the requirements and the code rather than adding
  a new obligation — worth a traceability pass either way.
- **Prior art on the shape.** The integration fixtures already model this from an earlier design
  generation: `tests/fixtures/integration/profile_owned_cli_state/expected/claude/composite-profile-summary.json`
  has `controls.environment` merged across inherited and selected profiles, with an adapter-specific
  `controls.claude.environment` layered on top, and the result in `launchEnv`. Those fixtures are
  stale in other respects, but the merge semantics they describe are a reasonable starting point.
- **Precedence has to be decided explicitly.** Inherited profile → selected profile → adapter-specific
  block is the fixture's order. Whether a profile value overrides an inherited *process* variable, or
  yields to it, is the real question — pi's `PI_SKIP_VERSION_CHECK` precedent in
  `resolveAgentLaunchExecutable` spreads the plan's env *under* `launchPlan.env`, so an existing
  convention exists to follow or to reject deliberately.

### Related requirement

OFTR-006.5.3, OFTR-006.3.2

### Harnesses affected

pi and Claude Code; Codex has the same need but no env projection today.

### What are you doing instead?

Exporting the variables in the shell, or wrapping `outfitter run` in a script — which puts the part
that makes the profile work outside the profile.

---

**One design constraint worth settling before implementation.** An environment block is an
exfiltration primitive: a profile that sets `ANTHROPIC_BASE_URL` redirects every prompt and
completion in the session to a host of its choosing, and one that sets `ANTHROPIC_AUTH_TOKEN` or
`HTTP_PROXY` is similar. That is completely reasonable from a profile the user wrote in `~/.agents`,
and not reasonable from a profile a remote catalog shipped them.

`isolation` faced the same question in #317 and answered it with `isHomeScope` gating in
`SettingsLoader.ts` — the setting is honored only from the user's own settings, so a checked-in
project or a remote catalog cannot decide how much of the user's machine a profile it ships can see.
Whatever this issue lands on should be a deliberate answer rather than an accident: scope-gate the
key, restrict which variable names a non-home-scope profile may set, or require confirmation for a
catalog profile that carries one. Silently honoring `env:` from any layer is the one outcome to
avoid.

luce-unsup:
## Summary
- add schema-validated `env` mappings to agent frontmatter and compose inherited values parent-first
- project trusted profile environment to Pi, Claude, and Codex launch plans
- honor environment declarations only from user-home agent definitions; project/source declarations warn and fail under `--strict`
- keep `PI_CODING_AGENT_DIR` and `CLAUDE_CONFIG_DIR` as Outfitter-owned boundaries while allowing profile values to override ordinary inherited process variables
- document merge, trust, secret-handling, and adapter behavior

## Validation
- `npm run check-ci` — 51 files / 582 tests passed before the final focused safety assertion
- `npx vitest --run tests/unit/profile-environment.test.ts` — 5 tests passed
- `npm run lint`
- `npm run build`

Fixes #319

### DIFF
--- a/code/cli/src/composer/Composer.ts
+++ b/code/cli/src/composer/Composer.ts
@@ -34,6 +34,7 @@ interface DeclaredSlug {
 
 interface EffectiveControls {
   readonly loadout: Loadout;
+  readonly environment: Readonly<Record<string, string>>;
   readonly skillSelections: readonly DeclaredSlug[];
   readonly subagentSelections: readonly DeclaredSlug[];
   readonly mcpSelections: readonly DeclaredSlug[];
@@ -165,6 +166,21 @@ const nearest = <T>(
   return undefined;
 };
 
+const composeEnvironment = (chain: readonly ChainEntry[], warnings?: string[]): Readonly<Record<string, string>> => {
+  const environment: Record<string, string> = {};
+  for (const entry of chain) {
+    if (Object.keys(entry.definition.environment).length === 0) continue;
+    if (entry.resource.winner.layer.origin !== 'global') {
+      warnings?.push(
+        `agent '${entry.resource.slug}' launch environment is ignored because only user-home agent definitions may control the process environment.`,
+      );
+      continue;
+    }
+    Object.assign(environment, entry.definition.environment);
+  }
+  return environment;
+};
+
 const composeTools = (chain: readonly ChainEntry[]): Loadout['tools'] => {
   const allow = union(
     [],
@@ -188,7 +204,7 @@ const nearestPrompt = (
   return undefined;
 };
 
-const composeEffectiveControls = (chain: readonly ChainEntry[]): EffectiveControls => {
+const composeEffectiveControls = (chain: readonly ChainEntry[], warnings?: string[]): EffectiveControls => {
   const skills = declaredSelections(chain, (definition) => definition.loadout.skills);
   const subagents = declaredSelections(chain, (definition) => definition.loadout.subagents);
   const mcp = declaredSelections(chain, (definition) => definition.loadout.mcp);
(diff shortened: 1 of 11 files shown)


ncrmro:
Closing in favor of #321. This pull request implements arbitrary launch environment. The replacement uses the resolved models.json registry and translates provider targets through each harness adapter.

PR #350, opened 2026-09-03
feat(settings): compose agent defaults into every agent

### LINKED ISSUE #347: feat(settings): compose agent defaults into every agent
## Problem

Organizations currently have no generalized `.agents`-wide way to add a skill, MCP server, plugin, subagent, prompt fragment, or extension to every agent. They must duplicate loadout entries in every `agents/<id>/agent.md` or force unrelated agents through a synthetic base profile.

The first concrete use case is **Pensieve**: an organization MUST be able to declare the Pensieve Pi collector once and have every Pi agent launched locally, in Actions, or as an Agent Operator resident receive that extension. This MUST remain a generalized composition primitive; Outfitter MUST NOT add Pensieve-specific settings or a `capture-profile` concept.

## Proposed shape

```yaml
agent_defaults:
  extensions:
    - git:github.com/ai-outfitter/pensieve@<immutable-ref>
  skills:
    - organization-practices
  mcp:
    - github
  append_system_prompt:
    - file: prompts/organization.md
```

`agent_defaults` composes into every selected agent before that agent's own loadout. `agents.md` remains shared prompt context rather than becoming a configuration manifest.

## Requirements

- `settings.yml` MUST accept an optional `agent_defaults` object.
- The first version MUST support only existing additive loadout fields: `extensions`, `skills`, `mcp`, `plugins`, `subagents`, and `append_system_prompt`.
- Composition MUST use the same deterministic parent-first ordering and stable de-duplication semantics as inherited agent loadouts.
- An agent's own loadout MUST compose after `agent_defaults`.
- Settings-layer precedence MUST remain deterministic and MUST be visible in validation/dump provenance.
- `outfitter run`, `outfitter dump`, and `outfitter validate` MUST operate on the same effective composition.
- Existing harness support rules MUST remain unchanged; for example, Pi loads `extensions`, while harnesses that cannot load Pi extensions do not report them as an actionable unsupported setting.
- The schema and implementation MUST remain backend-neutral and MUST NOT contain Pensieve-specific fields, sink URLs, credentials, retention policy, or workload-identity behavior.
- Existing catalogs without `agent_defaults` MUST retain identical behavior.

## Acceptance

A fixture catalog with two unrelated Pi agents declares the pinned Pensieve extension only once under `settings.yml:agent_defaults.extensions`. Validation succeeds, both dumped agents contain exactly one resolved Pensieve extension, and neither agent profile mentions Pensieve.

Tests also cover composition with agent-specific loadout entries, duplicate suppression, settings-layer precedence, missing referenced resources, and unchanged behavior when `agent_defaults` is absent.

## Out of scope

- Provisioning a Pensieve backend.
- Supplying sink endpoints or credentials.
- GitHub App webhook delivery or workload-token exchange.
- Agent Operator enforcement or managed collector installation.

Those execution environments consume the resolved declaration and provide their own runtime bindings.

luce-unsup:
## Summary

Implements #347: `settings.yml` MAY declare an optional `agent_defaults` object whose additive loadout entries compose into **every** agent — ahead of that agent's own loadout — so an organization declares a shared extension (the driving use case: a pinned Pensieve Pi collector), skill, MCP server, plugin, delegate, or appended prompt fragment once instead of duplicating it into every `agents/<id>/agent.md`.

## Behavior

- `agent_defaults` supports only the existing additive loadout fields: `extensions`, `skills`, `mcp`, `plugins`, `subagents`, and `append_system_prompt` (`file:`/`repo_file:`). Scalar overrides (`model`, `thinking`, `tools`) stay agent-owned; the schema rejects unknown keys.
- Composition uses the same deterministic parent-first ordering and stable de-duplication as inherited agent loadouts: defaults compose first (like a root-most ancestor), then the inheritance chain, selected child last. Duplicate slugs collapse to their first occurrence, so an agent's own entries never duplicate a default.
- Defaults resolve catalog-wide across layers — never through an agent's local namespace — and settings-layer prompt fragments resolve against the highest-precedence layer containing the file (`repo_file` keeps repository-trust machinery). Delegates are agents too: they receive the defaults as well.
- Unresolved references surface as composition warnings and `outfitter validate` findings named `agent_defaults …` (unresolved `skills`/`subagents` fail validation like agent-declared loadout refs; `mcp` stays a warning), so the settings layer is visible in validation.
- `outfitter run`, `outfitter dump`, and `outfitter validate` operate on the same effective composition (single `compose` path fed merged settings). Harness support rules are unchanged: defaults ride existing loadout projection, so non-Pi harnesses keep using the standard unsupported-element diagnostics.
- `outfitter dump` records `agentDefaults` provenance per composition in `.outfitter/composition.json` (fragments carry `declaringAgent: "settings:agent_defaults"`) and writes the merged defaults into the dumped tree's `settings.yml`, keeping the dump self-contained. Workflow dumps merge identically.
- Every loaded settings scope (remote, user, user-local, project, project-local) may declare defaults; fields merge as ordered-set unions collapsing cross-file duplicates to their first occurrence, lowest-precedence first.
- Settings without `agent_defaults` compose, validate, run, and dump exactly as before: plans gain no key, dumps write no `settings.yml`.

## Out of scope (per the issue)

No Pensieve-specific settings, sink URLs, credentials, retention policy, workload-identity behavior, or backend provisioning — the composition primitive stays backend-neutral, and `agents.md` remains shared prompt context.

## Testing

- New `tests/unit/agent-defaults.test.ts` (16 tests) covers the issue's acceptance end to end: the two-unrelated-agents fixture validates, dumps exactly one resolved Pensieve extension per agent with no mention in either profile, and re-resolves identically from the dumped tree — plus scope merging, ordering/dedup, catalog-wide resolution, unresolved-reference diagnostics, missing/optional prompt sources, delegate composition, and absent-defaults parity.
- Existing composer/dump/settings/resolver-validation suites pass unchanged; full suite 620 tests green with the coverage gate met.
- Manual smoke: built CLI against a fixture catalog — `validate` passes, `dump --agent` writes `settings.yml` + flattened prompt/skill + provenance, and the dumped tree re-validates standalone.

Closes #347

### DIFF
--- a/code/cli/src/dump/Dump.ts
+++ b/code/cli/src/dump/Dump.ts
@@ -127,6 +127,7 @@ interface CompositionProvenance {
 interface ClosureCompose {
   readonly agents: readonly ResolvedResource[];
   readonly skills: readonly ResolvedResource[];
+  readonly agentDefaultMcpServers: Readonly<Record<string, unknown>>;
   readonly promptFiles: readonly { readonly reference: string; readonly content: string }[];
   readonly provenance: readonly CompositionProvenance[];
   readonly warnings: readonly string[];
@@ -217,6 +218,16 @@ const collectSkills = (
   }
 };
 
+const collectAgentDefaultMcpServers = (
+  plan: CompositionPlan,
+  defaults: AgentDefaults | undefined,
+  servers: Record<string, unknown>,
+): void => {
+  for (const slug of defaults?.mcp ?? []) {
+    if (Object.hasOwn(plan.loadout.mcpServers, slug)) servers[slug] = plan.loadout.mcpServers[slug];
+  }
+};
+
 /** BFS over the selected agent and its delegated-agent closure. */
 const composeClosure = (
   set: EffectiveResourceSet,
@@ -229,6 +240,7 @@ const composeClosure = (
   const skills = new Map<string, ResolvedResource>();
   const warnings: string[] = [];
   const errors: string[] = [];
+  const agentDefaultMcpServers: Record<string, unknown> = {};
   const promptFiles = new Map<string, string>();
   const provenance: CompositionProvenance[] = [];
   const queue = [rootSlug];
@@ -254,6 +266,7 @@ const composeClosure = (
     collectContributingAgents(set, composed.plan.inheritanceChain!, agents);
     provenance.push(compositionProvenance(composed.plan));
     warnings.push(...composed.plan.warnings);
+    collectAgentDefaultMcpServers(composed.plan, agentDefaults, agentDefaultMcpServers);
     collectPromptFiles(composed.plan, promptFiles, errors);
     collectSkills(composed.plan.loadout.skills, skills, errors);
     queue.push(...composed.plan.loadout.subagents.map((subagent) => subagent.slug).sort(compareSlugs));
@@ -262,6 +275,7 @@ const composeClosure = (
   return {
     agents,
     skills: [...skills.values()].sort((left, right) => compareSlugs(left.slug, right.slug)),
+    agentDefaultMcpServers,
     promptFiles: [...promptFiles.entries()]
       .map(([reference, content]) => ({ reference, content }))
       .sort((left, right) => compareSlugs(left.reference, right.reference)),
@@ -316,6 +330,26 @@ const writeDefaultsSettings = (outRoot: string, defaults: AgentDefaults | undefi
   const settingsTarget = join(outRoot, 'settings.yml');
   writeFileSync(settingsTarget, settingsYaml);
   written.push(settingsTarget);
+};
+
+/** Makes settings-selected MCP servers resolvable from the flattened root copied into the dump. */
+const writeAgentDefaultMcpServers = (
+  outRoot: string,
+  effectiveServers: Readonly<Record<string, unknown>>,
+  written: string[],
+): void => {
+  if (Object.keys(effectiveServers).length === 0) return;
+
+  const target = join(outRoot, 'mcp.json');
+  let existingServers: Readonly<Record<string, unknown>> = {};
+  try {
+    const document = JSON.parse(readFileSync(target, 'utf8')) as { readonly mcpServers?: Record<string, unknown> };
+    existingServers = document.mcpServers ?? {};
+  } catch {
+    // Composition already reported malformed source MCP JSON; emit the effective valid subset.
+  }
+  writeFileSync(target, `${JSON.stringify({ mcpServers: { ...existingServers, ...effectiveServers } }, null, 2)}\n`);
+  written.push(target);
 };
 
 const failure = (errors: readonly string[], warnings: readonly string[] = []): DumpResult => ({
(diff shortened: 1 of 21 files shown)


code/cli/src/dump/Dump.ts
```diff
@@ -360,6 +400,10 @@ export const dumpAgent = (
   writeFileSync(provenanceTarget, `${JSON.stringify({ version: 1, compositions: closure.provenance }, null, 2)}\n`);
   written.push(provenanceTarget);
 
+  // Carry the merged settings-layer defaults into the dumped tree so it resolves identically on
+  // its own; catalogs without agent_defaults dump byte-identically to before.
+  writeDefaultsSettings(outRoot, agentDefaults, written);
```
ncrmro, line 439: [P2] Materialize the effective MCP definitions before carrying these defaults into the dump. When a lower-precedence layer defines an `agent_defaults.mcp` server and a higher-precedence layer has an `mcp.json` containing only unrelated servers, composition finds the lower definition across all root files. `writeRootFiles`, however, copies only the highest-precedence whole `mcp.json`, while this line preserves the selected slug in `settings.yml`. Reloading or running the supposedly self-contained dump then reports the server as unknown and loses part of the original composition. Write a merged MCP map containing the effective selected definitions, or otherwise ensure every emitted default remains resolvable.

ncrmro (changes requested):
The new agent-defaults dump path is not self-contained when a selected MCP definition comes from a lower-precedence layer. The focused test could not start because the repository's Nix shell currently fails while importing an npm lock entry without `integrity`; static review and five independent lanes otherwise covered the changed paths.

code/cli/src/dump/Dump.ts
ncrmro: Fixed in `ff8f41a`: dumps now carry the effective MCP definitions selected by `agent_defaults` into the flattened root `mcp.json`. Added round-trip coverage for a lower-layer definition shadowed by an unrelated higher-layer file, plus the missing-map fallback. Full `npm run check-ci` passes (58 files, 649 tests, branch coverage 98.01%).

ncrmro (approved):
Verified the MCP dump round-trip fix at `ff8f41a`. The effective `agent_defaults.mcp` definitions are now retained in self-contained dumps, the regression cases pass, and the full CI-equivalent gate is green.

code/cli/src/dump/Dump.ts
luce-unsup: Confirmed this addresses the P2 finding. `composeClosure` now collects the effective definitions for every slug selected by `agent_defaults.mcp` from the composed loadout map (so lower-precedence layers are covered), and `writeAgentDefaultMcpServers` merges them into the dumped root `mcp.json` after `writeRootFiles` copies the highest-precedence document (ordering verified: writeRootFiles at line 426, merge at line 440), with the effective selected definition winning on slug conflicts.

The two new focused tests cover the exact reported scenario (home-layer `github` selected from project-layer settings alongside an unrelated project `slack`) and assert the dump is self-contained via a fresh `resolveEffectiveSet` + `compose` reload with zero warnings, plus the empty-server-map replacement case. CI is green on `ff8f41a` across all four lanes (ubuntu/macos/windows Check + packaged E2E smoke).

Thank you for pushing the fix — re-requesting your review.

PR #379, opened 2026-09-07
feat(workflows): resolve output types as catalog resources

ncrmro:
Stacked on #378 (base `feat/workflow-outputs`). Follows @ncrmro's question on #378: with the type enum and value schemas built into the CLI, Outfitter accumulates one schema per type and every new type is a release. This PR makes output types a catalog resource kind instead.

## What changes

- **`output-type` resource kind** at `output-types/<slug>/schema.json`, top-level, resolved with the same layer precedence and shadow reporting as agents, skills, and workflows. `list output-types [--json]` lists them; `DefaultCatalog` recognizes the directory.
- **Workflow `type` is a slug** that MUST resolve to an output-type resource. The enum and the four built-in `output-type.*.schema.json` files are gone; the community catalog publishes `pull-request`, `git-commit`, `git-branch`, `issue` (ai-outfitter/community-profiles#83). An unresolvable type, an invalid schema, or a schema without an absolute-URI `$id` is a hard error, reported once per resource.
- **Identity is `($id, sha256)`, never the bare slug.** `workflow-composition.json` gains `outputTypes: [{slug, id, sha256, source}]`; every resolved output entry (manifest and `list workflows --json`) carries `schema: <$id>`; the resolved schemas are copied verbatim into the bundle at `.agents/output-types/<slug>/schema.json`, so their `files` hash equals the recorded digest and a consumer can validate offline. A shadowed output type warns, fatal under `--strict`.
- **`validateOutputValue(schema, value)`** compiles a caller-supplied schema (cached) rather than a built-in.
- **OFTR-013 amended** per OFTR-008: 013.3.1 and 013.3.2 replaced in place with `REQUIREMENT REMOVED`, new statements appended for resolution, identity, export, and the A2A `schema`/`digest` keys (mirrored in ai-outfitter/channels#69).

## Why identity, not just provenance

An independent Codex Sol design review preferred keeping the enum in the CLI while a type was identified by a bare name, because two catalogs could define `git-commit` differently and a cross-repository prerequisite could not compare them. It accepted the catalog model once identity became `($id, digest)` carried in the manifest and the A2A artifact. That amendment is the last commit here.

## Verification

- `OUTFITTER_SYSTEM_DIR=<empty dir> npm run check-ci`: 753 tests, 98.06% branch coverage; `npm run build` passes (checked separately since vitest does not typecheck).
- Output-type schemas may use standard JSON Schema formats (`ajv-formats`, added with an OFTR-001.5 amendment); an unknown format or a cross-schema `$ref` is a schema defect at validate and dump time, not a consumer-time throw.
- Community catalog with `output-types/` added: `scripts/validate-workflows.sh` against this build passes strict validation and byte-identical double export for all ten roots; the `issue-triage` bundle carries both used schemas, and the manifest records their `$id`, digest, and winning layer.
- Adversarial Claude review (read-only, opus): `0013d39` changes requested (missing `escapesRoots` containment on the schema copy, shared-Ajv `$id` collision throw, shadow-warning attribution, vacuous sort test); `120ea96` fixed those but compiled schemas without standard formats; `db7b575` approved with two advisories, applied in `373c8b7`.

Merge order: this PR into `feat/workflow-outputs` first (merge or squash into the branch), then #378 into main as one squash.


ncrmro:
Closing unmerged per #380. Two independent design reviews and the owner converged on keeping `type` a plain label in Outfitter: nothing consumes a value schema yet, removing machinery later would be breaking while adding it is additive, and the `($id, sha256-of-bytes)` identity here forks on a reformat. #378 is being trimmed to labels instead.

What survives for a future reopen is recorded on #380: resource-kind resolution with layer precedence was sound; the corrections from review (canonical-JSON digest, self-contained schemas, containment on copies, default-catalog pin sequenced first, no validator without a caller) are the starting point if two engines ever exchange values. The branch stays until #378 lands.

PR #394, opened 2026-09-08
fix(sync): defer transitive inheritance resolution

### LINKED ISSUE #393: fix(sync): defer inherited parents until transitive catalogs resolve
## Problem

`outfitter sync` validates each directly configured remote source in isolation before discovering that source's transitive `sources` closure. An agent that inherits a parent supplied by a declared transitive catalog therefore fails the initial source fetch with `Agent inheritance references unknown parent`, and dependency discovery never runs.

This contradicts the intended transitive catalog composition. A concrete deployment has:

```yaml
# generated ~/.agents/settings.yml
sources:
  - uri: git+https://git.example.test/example/.agents.git
    ref: <full commit SHA>
    path: .
  - path: /opt/agent/.agents
```

The fetched catalog contains:

```yaml
# agents/luce/agent.md
inherits: [engineer, environment.agent-operator-pod]
```

and:

```yaml
# settings.yml in that fetched catalog
sources:
  - github: ai-outfitter/community-profiles
    ref: v1.8.0
```

Outfitter 1.15.0 exits before composition with:

```text
Source-cache repair policy failed before composition: failed: source ...
(Fetched source is invalid: agent:luce Agent inheritance references unknown parent 'engineer' ...;
agent:luce Agent inheritance references unknown parent 'environment.agent-operator-pod' ...)
```

Flattening the child agent makes the source sync, which confirms transport, authentication, and the pinned catalog revision are healthy, but it defeats reusable catalog inheritance.

## Root cause

`validateSourceCheckout` in `SyncCommand.ts` calls:

```ts
validateEffectiveSet(resolveResources([layer]), undefined, {
  deferLoadoutResolution: true,
});
```

That correctly defers unresolved skill/subagent loadout references while a source is isolated, but `validateAgent` always reports composer errors for unknown inherited parents as fatal. The source is rejected before `syncTransitiveClosure` can read its `settings.yml` and fetch `community-profiles`.

## Expected behavior

During isolated source validation, an unresolved inherited parent MUST be deferred just like an unresolved loadout agent/skill when it may be supplied by a declared transitive source. After the dependency closure is fetched, normal merged-set validation MUST still reject a genuinely missing parent.

## Suggested regression test

Extend `code/cli/tests/unit/sync-loadout-deferral.test.ts` with two fixture repositories:

1. A top catalog containing `child` with `inherits: [parent]` and a pinned `github:` dependency.
2. The dependency catalog containing abstract agent `parent`.

`executeSyncCommand` should return exit code 0 and report one direct source plus one transitive source as updated. Existing authoritative merged validation should continue to fail if `parent` is absent after closure resolution.

## Relevant requirement

OFTR-004.6 establishes transitive source discovery and already requires isolated validation to defer unresolved references that may be supplied by a catalog dependency. The requirement should explicitly include inherited parents so validation cannot deadlock dependency discovery.

ncrmro:
Fixes #393.

## Summary

- defer unresolved inherited parents while validating an individual catalog source during sync
- retain fatal validation for inheritance cycles, malformed inheritance, unrelated composition errors, and unresolved parents in the final merged catalog
- cover direct agents, subagents, workflow actors, transitive fetching, and final `base -> child` composition
- extend OFTR-004.6.11 with the transitive inheritance contract

## Verification

- focused resolver/sync tests: 7/7 pass
- ESLint: pass
- Prettier: pass
- two independent read-only review lanes: approve
- `npm run check-ci`: formatting and lint pass; 746/747 tests pass. The sole failure is the pre-existing `tests/unit/setup.test.ts` EISDIR failure also observed before this change.

### DIFF
--- a/code/cli/src/cli/commands/SyncCommand.ts
+++ b/code/cli/src/cli/commands/SyncCommand.ts
@@ -118,10 +118,13 @@ const validateSourceCheckout = (repositoryPath: string, source: RemoteSourceRefe
     throw new Error(`Configured source path '${source.path}' was not found in the fetched repository.`);
   }
 
-  // A source is validated in isolation here, so an unresolved loadout slug is deferred (a warning,
-  // not a failure): a catalog may reference a skill supplied by a catalog it declares as a transitive
-  // dependency (OFTR-004.6), and loadout wholeness is validated against the merged tree at resolution.
-  const findings = validateEffectiveSet(resolveResources([layer]), undefined, { deferLoadoutResolution: true });
+  // A source is validated in isolation here, so unresolved loadout slugs and inheritance parents
+  // are deferred (warnings, not failures): a catalog may reference resources supplied by a catalog
+  // it declares as a transitive dependency (OFTR-004.6). The merged tree remains authoritative.
+  const findings = validateEffectiveSet(resolveResources([layer]), undefined, {
+    deferLoadoutResolution: true,
+    deferInheritanceResolution: true,
+  });
   const errors = findings.filter((finding) => finding.severity === 'error');
   if (errors.length > 0) {
     throw new Error(
--- a/code/cli/src/resolver/ResolverValidation.ts
+++ b/code/cli/src/resolver/ResolverValidation.ts
@@ -15,15 +15,15 @@ export interface ValidationFinding {
 }
 
 /**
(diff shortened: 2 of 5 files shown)


ncrmro (commented):
APPROVE — two independent read-only review lanes approved amended commit b906504. The review verified direct, subagent, and workflow-actor deferral; final merged validation; cycle/error preservation; and transitive base -> child composition coverage.

PR #409, opened 2026-09-14
fix(resolver): collapse workspace and global layers when they are the same directory

### LINKED ISSUE #336: fix(resolver): deduplicate workspace and global layers when cwd is home directory (~)
### Problem Statement

When executing `outfitter list` or `outfitter validate` from the user's home directory (`$HOME` / `~`), Outfitter resolves both:
1. Workspace layer: `<cwd>/.agents` (`/home/user/.agents`)
2. Global layer: `~/.agents` (`/home/user/.agents`)

Because both layers point to the exact same directory on disk, Outfitter sees identical agent slugs across both layers and emits false-positive ambiguity and shadowing warnings:

```
⚠ settings: Ambiguous agent slug 'coding' is supplied by 'workspace', 'global'; 'workspace' won.
⚠ agent:coding: shadowed definition in global is overridden by workspace.
```

### Proposed Solution

In the layer resolver (`Resolver.ts` / layer collection logic):
- Compare the canonical real paths of the workspace root and the global root (`realpath(workspaceRoot) === realpath(globalRoot)`).
- If they resolve to the identical directory, collapse them into a single global layer.

### Benefits
- Eliminates false-positive warnings when running Outfitter commands from the home directory.
- Clearer diagnostics for real shadowing across distinct project and user layers.

raphaelbahat:
## Summary

When running `outfitter list`, `outfitter validate`, or any other effective-set command from the user's home directory, layer discovery resolved both the workspace layer (`<cwd>/.agents`) and the global layer (`~/.agents`) to the exact same directory. Every resource was discovered twice, producing false-positive diagnostics:

```
⚠ settings: Ambiguous agent slug 'coding' is supplied by 'workspace', 'global'; 'workspace' won.
⚠ agent:coding: shadowed definition in global is overridden by workspace.
```

`discoverLayers` now compares the canonical real paths of the two roots (error-tolerant `realpathSync`) and skips the workspace candidate when they are identical, keeping one **global** layer. Because the fix lives in the single shared resolver path, `list`, `validate`, `run`, `dump`, `link`, and `sync` all benefit — no per-command patches.

## Before / after

- **Before:** from `~`, resources under `~/.agents` appear twice (workspace + global) and emit ambiguity/shadowing warnings naming both layers; agent `configPaths`/`mcpPaths`/`piConfigDirectories` contain duplicate entries of the same absolute path.
- **After:** from `~`, one global layer; no self-shadowing warnings; winners, slugs, and loadouts identical (the duplicated root resolved to identical files, so dropping the duplicate cannot change outcomes).
- **Unchanged:** distinct roots (e.g. cwd `~/projects`) still collect workspace over global, still shadow, and still warn exactly as today; missing roots are still filtered by `existsSync` without crashing.

## Collapse semantics and layer-identity decision

- The surviving layer is `{ origin: 'global', label: 'global' }`. Evidence: warnings name layers via `Layer.label` (`AmbiguityWarnings.ts`), and `docs/documentation/concepts.md` defines `~/.agents` as the global layer. Attributing `~/.agents` content to `global` keeps diagnostics stable across working directories — from any other cwd the same files already report as `global`. Keeping `workspace` would make attribution cwd-dependent, and precedence order only exists to break ties between *distinct* roots; with one root there is no tie.
- Symlink equality counts in both directions (`~/.agents -> <project>/.agents` and the reverse) because `realpathSync` canonicalizes through links.
- Unresolvable paths (missing directory, `EACCES`) never collapse: discovery observes identical behavior for missing roots as before, and the existing `existsSync` filter still decides inclusion.
- Settings scopes are untouched: `SettingsLoader` reads user/project scopes through its own path list and keeps its existing precedence.

## Test evidence

New focused suite `code/cli/tests/unit/resolver-home-collapse.test.ts` (temp-dir HOME fixtures, no real `/home/bahat`):

- home-cwd resolution collapses to one `global` layer with no ambiguity/shadowing warnings and the same winner/path as before
- symlinked root equality collapses in both directions
- distinct roots still warn and still resolve workspace-over-global (real shadowing preserved)
- missing payload roots do not crash discovery
- every shared-resolver consumer (`resolveEffectiveSet`, `list`, `validate` incl. `--strict`) is warning-free from home
- settings scopes remain readable when the directory coincides (path sources still become layers)

Checks: `npx prettier --check .` clean (only the local-only `.pi/tasks` fixture warns, as on main), `npm run lint` clean, `npm run coverage` 755 tests passing with thresholds met. (One pre-existing environmental flake: `source-cache-policy.test.ts` can time out under parallel machine load; it passes in isolation and on `main`.)

Docs: `docs/documentation/concepts.md` (Layers), `docs/architecture/README.md` (resolution semantics), and a new formal item OFTR-003.3.4 in `docs/requirements/OFTR-003-profiles.md`.

Fixes #336


raphaelbahat:
Follow-up commit 69c17cc (pushed to this branch): the conditional spread in the candidates array broke contextual typing, widening the workspace layer's `origin` to `string` — `tsc -p tsconfig.build.json` (the build, not vitest) caught it. The candidate is now hoisted into a `Layer`-typed constant; no runtime behavior change, full suite (755/755) and coverage gates still green.

PR #404, opened 2026-09-14
fix(pi-adapter): preserve foreign subagent files across the rebuild

raphaelbahat:
## Summary

`outfitter run --harness pi` rebuilt the runtime `agents/` directory wholesale whenever the selected profile declared `subagents:` delegates, silently wiping agent definitions delivered through the documented `agents/<agent>/pi/` configuration overlay (OFTR-006.3.17) in the same run.

This change replaces the wholesale removal with a **manifest-scoped cleanup semantic**: subagent materialization records the delegate files it generated in a deterministic rebuild manifest under the projection root, and a later projection into the same root removes only manifest-tracked regular files that stay inside the root.

## Reproduction → behavior

- **Before:** with a `agents/<agent>/pi/agents/general-purpose.md` overlay and a profile declaring `subagents:`, the runtime `agents/` contained only the declared delegates — the overlay file was gone (reproduction from #401).
- **After:** the overlay file survives every run. Declared delegates are still written, a declared delegate wins a same-slug collision with overlay content, and shrinking the delegate loadout (including to zero) still cleans stale Outfitter-generated delegates in a reused/retained root. A missing, corrupt, or hostile manifest disables cleanup instead of risking foreign deletions; cleanup never removes symlinks, directories, or anything outside the root.

The projection ordering (pi/ overlay before materialization) is unchanged, preserving declared-delegate authority — the alternative of moving the overlay after materialization was rejected in the issue because it silently inverts precedence. The manifest tracks exactly the files Outfitter's subagent materialization generated, so files from any other delivery mechanism (including the planned settings-layer overlay in #402) count as foreign and survive without further changes.

## Test evidence

- New focused suite `code/cli/tests/unit/subagent-rebuild-manifest.test.ts` (16 tests): overlay-only survival, survival when delegates are declared, stale generated-delegate cleanup after shrink and shrink-to-zero, collision precedence, manifest determinism (byte-identical, sorted relative paths, no timestamps), foreign files never tracked, corrupt/structurally-invalid manifest robustness, out-of-root/symlink/directory manifest paths ignored, vanished tracked files tolerated, skipped subagents not tracked, and an end-to-end run-command regression reading the projection at launch time.
- Full suite: 765 tests pass; `npm run check` and `npm run check-ci` both pass (prettier, ESLint, coverage thresholds).

## Docs

- `docs/requirements/OFTR-006-agent-adapters.md`: new OFTR-006.3 item 21 — the manifest-scoped rebuild ownership contract.
[...]

PR #414, opened 2026-09-15
feat(launcher): add an opt-in to launch PATH pi via OUTFITTER_PI_BIN and pi_binary settings

### LINKED ISSUE #345: feat(launcher): pin bundled pi to current 0.84.x and add an opt-in to launch PATH pi (OUTFITTER_PI_BIN)
### Problem Statement

The pi adapter launches the **bundled** `@earendil-works/pi-coding-agent` resolved from Outfitter's own dependency closure (`AgentLaunch.ts` → `resolveBundledPiBinPath()`), which is correct for reproducibility — but the dependency is currently pinned to `"^0.80.3"` while pi.dev's latest is **0.84.x**. Every `outfitter run --harness pi` therefore runs a Pi version several minor releases behind the user's standalone `pi`.

This has real consequences:

1. **Version skew between launch modes** — `outfitter run` (bundled 0.80.10) vs raw `pi` (0.84.4) behave differently (new tools, fixes, skill/extension semantics) for the same profile.
2. **Extension-cache installs run old-Pi semantics** — `PiExtensionCache` spawns `pi install` against the cache agent dir, so cache-time behavior is also the bundled version's.
3. The launcher itself documents the gap (`AgentLaunch.ts`): *"the pinned pi can still lag pi.dev's latest"* — and forces `PI_SKIP_VERSION_CHECK=1` so users are never even told they're on a stale Pi.

### Proposed Solution (both parts)

1. **Bump the bundled dependency** to `@earendil-works/pi-coding-agent@^0.84` (current latest 0.84.4), restoring near-parity with pi.dev's latest for bundled launches.
2. **Add an opt-in to launch the user's PATH `pi`**, e.g.:
   - an `OUTFITTER_PI_BIN` environment variable, and/or
   - a settings key (e.g. `pi_binary: path | bundled | auto` in `settings.yml`),
   with `bundled` remaining the default. `resolveBundledPiLaunch()` already has a natural fallback seam: when bundled resolution fails it falls through to a PATH lookup — the opt-in would make that PATH path selectable rather than exceptional.

### Verification performed

- Confirmed the launch mechanics in `AgentLaunch.ts` (`resolveBundledPiLaunch` → `process.execPath` + bundled bin, `PI_SKIP_VERSION_CHECK: '1'`).
- Confirmed Outfitter's `dist/` never imports pi's JS API (bin-resolution only), so the bundled version is decoupled from adapter compilation.
- Confirmed no existing issue tracks this (searches for "bundled", "pi version", "pi-coding-agent", "PATH pi").
- Empirically: swapping the nested bundled package for 0.84.4 launches and runs correctly through `outfitter run <agent> -- --version` → `0.84.4`, i.e. the adapter works against 0.84.x today.

### Benefits
- Eliminates the silent version skew between Outfitter-managed and standalone Pi sessions.
- Gives users on fast-moving Pi releases a supported escape hatch instead of manual `node_modules` surgery.

raphaelbahat:
## Summary

Part 2 of #345: an opt-in to launch the user's PATH `pi` instead of the bundled one, with `bundled` remaining the default. Part 1 (#413) bumped the bundled `@earendil-works/pi-coding-agent` to `^0.85.1`.

## Design (confirmed in issue discussion)

- **Both surfaces**: settings keys for durable catalog config, plus the `OUTFITTER_PI_BIN` environment variable for one-off overrides.
- **Precedence**: `OUTFITTER_PI_BIN` (non-empty; a path value) > merged settings > built-in default `bundled`.
- **Key shape**: two keys — `pi_binary: bundled | path | auto` (enum, default `bundled`) and `pi_binary_path: <path>` (non-empty string, resolved relative to the declaring settings file, honored from every scope). Chosen over a single `bundled|path|<path>` union string because the enum + string validate unambiguously at the JSON Schema read boundary, while a union makes `pi_binary: bundled` grammatically ambiguous (mode or relative path?). `pi_binary_path` without `pi_binary` implies `path` mode.
- **`auto` semantics**: bundled first, PATH `pi` fallback when bundled resolution fails — formalizing `resolveBundledPiLaunch()`'s existing exceptional fallback seam rather than flipping its direction (the issue itself calls bundled resolution "correct for reproducibility"). The fallback emits a warning that is fatal under `--strict`.
- **Missing explicit binary**: `OUTFITTER_PI_BIN` or `pi_binary_path` (path mode) naming a nonexistent file fails the run before launch with an actionable error — a silent fallback to bundled would reintroduce, invisibly, the exact version skew the user opted out of.
- **`PI_SKIP_VERSION_CHECK`**: injected `'1'` only for bundled launches (its rationale — `pi update` cannot update a dependency-closure copy — is ownership-specific). User-selected binaries keep pi's own actionable update notice; a value already in the launch environment always wins.
- **Cache boundary**: `PiExtensionCache` installs keep spawning the bundled binary regardless of selection (the cache's spawner passes no selection; a test pins that a cache-style no-selection launch stays bundled even with `OUTFITTER_PI_BIN` set). Changing cache-time semantics remains out of scope.

## Implementation

- New `code/cli/src/agents/PiBinarySelection.ts`: pure selection resolution (env + settings in, selection + warnings/error out; the bundled-resolution probe is injectable so `auto` is testable), now also owning the bundled-bin resolution moved out of `AgentLaunch.ts` — one module owns "which pi binary".
- `AgentLaunch.resolveAgentLaunchExecutable(plan, selection?)` / `launchThroughSpawn(spawn, plan, selection?)` apply the selection exactly once at the spawn boundary; the default selection is byte-identical to the previous behavior.
- `RunAgentCommand` resolves the selection pi-scoped (non-pi harnesses never consult it), surfaces selection warnings into the existing strict-fatal warning set, and throws the pre-launch error before any spawn.
- Settings: `piBinary`/`piBinaryPath` leaves with leaf-wins merge; JSON Schema enum + minLength validated at the read boundary.

Stacked on #413 (⬅ #412 ⬅ #411 ⬅ #410 ⬅ #408 ⬅ #407 ⬅ #406 ⬅ #405 ⬅ #404)

## Test evidence

- 917 tests pass across 79 files (`npm run coverage`); statements 99.25%, branches 98.08%, functions 99.03%, lines 99.54% — above gates.
- New suites: `pi-binary-selection.test.ts` (resolution: precedence, modes, missing-binary error, ignored-path warning), `run-agent-pi-binary.test.ts` (command threading, strict-fatal warnings, pre-launch failure, env override, non-pi scoping, and an end-to-end real-spawn test launching a fixture binary that proves the selected process runs and receives no injected version-check suppression), plus extensions to `agent-launch.test.ts` and `settings-loader.test.ts` (schema validation, declaring-file path resolution, cross-layer precedence) with OFTR-002.12 / OFTR-006.3 traceability.
- `npx prettier --check .` clean (only the local-only untracked `.pi/` artifact warns), `npm run lint` clean, `npm run typecheck` shows only the two pre-existing errors.

Fixes #345 (together with #413)


raphaelbahat:
## Bugfix: settings-resolved paths now expand a leading `~`

**User-reported repro.** With `pi_binary_path: ~/.bun/bin/pi` in `~/.agents/settings.yml`, launch failed with:

```
Could not launch pi: 'pi_binary_path' names '/home/bahat/.agents/~/.bun/bin/pi', which does not exist on disk.
```

The leading tilde was joined against the declaring settings file's directory instead of expanding to the home directory.

**Systemic finding.** This was not specific to `pi_binary_path`: no settings-resolved path expanded `~` anywhere. `cache_directory: ~/x`, `sources: - path: ~/x`, and `agent_defaults.pi_overlay: ~/x` all shared the same latent trap — relative values were joined against the declaring settings directory with no home expansion.

**Fix shape (central, DRY).** All four surfaces route through the shared `resolveConfigDirectory` utility in `src/settings/SettingsLoader.ts`, so the expansion lives there once:

- A leading bare `~` or `~/...` expands to the user's home directory (POSIX convention). A `~name/...` form is **not** expanded and keeps the existing declaring-directory handling; already-absolute paths are untouched; plain relative values keep resolving against the declaring settings directory.
- The expansion keys on the run's resolved home directory input (`homeDirectory`), not the declaring settings directory — so project-scope `~/x` also expands to `$HOME/x`, never `<project-settings-dir>/~/x`.
- Plumbing: the home directory is stamped onto `SettingsLoadPlan` at discovery time (both local and cached-remote discovery paths already receive it) and threaded explicitly through `addSettingsFile` → `convertSettingsDocument` → the resolution site. No hidden global state, no `process.env.HOME` reads inside the loader.

**Regression coverage** (new focused suite `tests/unit/settings-path-tilde-expansion.test.ts` plus an end-to-end case in `run-agent-pi-binary.test.ts`):

- `pi_binary_path: ~/...` in home-scope settings resolves to `$HOME/...` — the reported bug.
- End-to-end through the real spawn boundary: home-scope settings declaring `pi_binary_path: ~/bin/fake-pi` launch the fixture binary successfully (the previous code failed pre-launch).
- Relative `pi_binary_path` still resolves against the declaring settings directory; absolute paths are untouched.
- `cache_directory`, source `path:`, and `agent_defaults.pi_overlay` tilde paths all expand to `$HOME/...`.
- A project-scope settings file declaring `pi_binary_path: ~/...` / `cache_directory: ~/...` resolves against the **user home**, not the declaring directory (pins the distinction from the declaring-directory-relative rule).
- A bare `~` value expands to the home root; a `~name/...` value is left unexpanded.
- Cross-layer precedence holds when user and project layers both declare tilde paths (higher layer wins, both expanded); env-var override precedence is unchanged.
- Requirements and docs updated to state the rule once at the shared-path level (`docs/requirements/OFTR-002-settings.md` clauses OFTR-002.5.3, OFTR-002.7.2, OFTR-002.10.12, OFTR-002.12.2; `docs/documentation/settings.md`; `docs/architecture/README.md`).

Verification: `npm run build`, `npm run typecheck` (only the two pre-existing test-file errors), `npx prettier --check .`, `npm run lint`, and the full coverage suite all pass.

